PR #4, opened 2021-07-22
add support for process events telemetry

RafaelOrtizRC:
Gathers telemetry information from the following syscalls:

* execve, execveat
    * command line args, environment variables, executable information, exit status
* fork
    * returned pid
* vfork
    * returned pid
* clone
    * clone args and returned tids
* clone3
    * clone3 specific argument struct and returned tids
* unshare
     * unshare flags
* exit, exit_group
    * exit status

There are versions for befor/after 4.8 without/with additional telemetry using struct offsets. `clone3` will only work on 5.3+.


src/types.h
```diff
@@ -46,14 +49,40 @@ typedef enum
...
+    SP_CLONE3,
+    SP_FORK,
+    SP_VFORK,
+    SP_EXECVE,
+    SP_EXECVEAT,
+} syscall_pattern_type_t;
+
+typedef enum
+{
+    Unspec,
```
rc-grey, line 65: can this be TE_UNSPEC to match convention

src/programs.c
```diff
@@ -181,11 +272,61 @@ static __always_inline syscall_pattern_t ptrace_syscall_pattern(u32 request)
         .mono_ns = mono_ns,                    \
     }
 
+#define FILL_TELEMETRY_SYSCALL_EVENT(E, SP)    \
```
rc-grey, line 275: Can the `\` be all at the same indent please

src/programs.c
```diff
@@ -181,11 +272,61 @@ static __always_inline syscall_pattern_t ptrace_syscall_pattern(u32 request)
...
+    bpf_get_current_comm(E->u.syscall_info.comm, sizeof(E->u.syscall_info.comm));
+
+#define FILL_TELEMETRY_SYSCALL_RET(E, SP) \
+
+
+#define GET_OFFSETS_4_8 \
+    /* if "loaded" doesn't exist in the map, we get NULL back and won't read from offsets                           \
+     * when offsets are loaded into the offsets map, "loaded" should be given any value                             \
+     */                                                                                                             \
+    u64 loaded = 0xec6642829d632573; /* CRC64 of "loaded"  */                                                       \
```
rc-grey, line 296: can we move the CRC constants to their own header file? that way we can autogenerate that header file in the future

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+
+static __always_inline void push_telemetry_event(ptelemetry_event_t ev)
+{
+    u32 key = 0;
+    u32 *pcurrent_index = bpf_map_lookup_elem(&telemetry_index, &key);
+    if (NULL == pcurrent_index)
+    {
+        return;
+    }
+    u32 idx = *pcurrent_index + 0;
```
rc-grey, line 797: whats the point of the `+ 0`?

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    if (ii < *pcurrent_index) {                                            \
+        ev = bpf_map_lookup_elem(&telemetry_stack, &ii);                     \
+        if (ev) {                                                           \
+            bpf_perf_event_output(ctx,                                      \
+                &telemetry_events,                                          \
+                bpf_get_smp_processor_id(),                                 \
+                ev,                                                         \
+                sizeof(*ev)                                                 \
+            );                                                              \
+            ii++;                                                        \
```
rc-grey, line 836: don't we always want to increment ii?

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+        }                                                                   \
+        else                                                                \
```
rc-grey, line 838: this can probably go away too?

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    u32 *pii = bpf_map_lookup_elem(&read_flush_index, &key);
+    if (NULL == pii)
+    {
+        bpf_map_update_elem(&read_flush_index, &key, &key, BPF_ANY);
+        return;
+    }
+
+    u32 ii = *pii;
+    ptelemetry_event_t ev = NULL;
+    FLUSH_LOOP
```
rc-grey, line 862: For both these and the read_loop lets adopt this pattern and put them in a header file.

```
#define REPEAT_0(WHAT)
#define REPEAT_1(WHAT) WHAT REPEAT_0(WHAT)
#define REPEAT_2(WHAT) WHAT REPEAT_1(WHAT)
#define REPEAT_3(WHAT) WHAT REPEAT_2(WHAT)

#define NOP_N(N) REPEAT_##N(asm("nop");)
```

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    FLUSH_LOOP
+    FLUSH_LOOP
+    FLUSH_LOOP
+    FLUSH_LOOP
+    bpf_map_update_elem(&read_flush_index, &key, &ii, BPF_ANY);
+}
+
+#define READ_LOOP(PTR, T)                                                       \
+    ev->id = id;                                                                \
```
rc-grey, line 899: Do the ev assignments need to happen every loop unroll? it seems like we are resusing the same memory space

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    u32 pid = pid_tgid >> 32;
+    ev->u.syscall_info.ppid = ppid;
+    ev->u.syscall_info.luid = luid;
+    push_telemetry_event(ev);
+
+    bpf_map_update_elem(&telemetry_ids, &pid, &id, BPF_ANY);
+
+
+    // explicit null check to satisfy the verifier
+    if (filename)
```
rc-grey, line 962: prefer `if (!filename) goto Exit` less indentation

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    return 0;
+}
+
+static __always_inline int exit_clone(struct pt_regs *ctx)
+{
+
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    u32 pid = pid_tgid >> 32;
+    u64 *id = bpf_map_lookup_elem(&telemetry_ids, &pid);
+    if (id)
```
rc-grey, line 1282: `if (!id) goto Exit;`

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    telemetry_event_t sev = {0};
+    __builtin_memcpy(&sev, pev, sizeof(telemetry_event_t));
+    ptelemetry_event_t ev = &sev;
+
+
+    id = bpf_get_prandom_u32();
+    ev->id = id;
+    ev->done = FALSE;
+    ev->telemetry_type = 0,
+            ev->u.v.value[0] = '\0';
```
rc-grey, line 1677: indent

src/programs.c
```diff
@@ -637,5 +778,1001 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+
+        ev->telemetry_type = TE_RETCODE;
+        ev->u.retcode = (u32)PT_REGS_RC(ctx);
+        push_telemetry_event(ev);
+
+    }
+    else
+        goto Flush;
+
+    Flush:
```
rc-grey, line 1724: indent

rc-grey (changes requested):

src/programs.c
RafaelOrtizRC: good idea

src/programs.c
RafaelOrtizRC: the verifier!

src/programs.c
RafaelOrtizRC: That's leftover from when `push_telemetry_event()` would return a fresh event from the map/stack, I'll move these outside the loop so we save a few instructions.

src/programs.c
RafaelOrtizRC: There _shouldn't_ be a case where we get a null pointer back from `bpf_map_lookup_elem` and there are events past it on the stack, so with `ii++` inside or outside the `if` we'll just keep skipping down the loop until we get to the end. It also shouldn't make a difference to move ie outside the `if (ev)` either and it would catch events if we accidentally skipped an element somehow.

src/programs.c
RafaelOrtizRC: yep - that's a leftover

src/programs.c
RafaelOrtizRC: 750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: 750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: 750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: Got it here and the same pattern in a few (all?) other places. 

750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: 750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: 750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: Also cleaned up some old/unused `Exit` labels.

750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: 750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: Added an `offsets.h` file with these values as constants.

750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
RafaelOrtizRC: Yes they can.

750d697f2242e69233ef10fb19251be28b8c7bae

src/types.h
RafaelOrtizRC: 750d697f2242e69233ef10fb19251be28b8c7bae

src/programs.c
```diff
@@ -637,5 +776,921 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    else if ((void *) ptr == NULL)                                              \
+    {                                                                           \
+        goto Next;                                                              \
+    }                                                                           \
+    else                                                                        \
+    {                                                                           \
+        count = bpf_probe_read_str(&ev->u.v.value, VALUE_SIZE, (void *) ptr);   \
+        if (count == VALUE_SIZE) {                                              \
+            ev->u.v.truncated = TRUE;                                           \
+        }                                                                       \
```
rc-grey, line 879:
``` else if (count > 0) {
    ev->u.v.truncated = FALSE;
} else {
    goto Next;
}```

File: src/programs.c [hunk 14]
```diff
@@ -637,5 +776,921 @@ int uprobe__read_return_string(struct pt_regs *ctx)
...
+    ev->u.v.truncated = FALSE;
+    long count = 0;
+    count = bpf_probe_read_str(&ev->u.v.value, VALUE_SIZE, (void*) filename);
+    if (count == VALUE_SIZE) {
+        ev->u.v.truncated = TRUE;
+    }
+    push_telemetry_event(ev);
+
+
+    bpf_tail_call(ctx, &tail_call_table, 0);
```
rc-grey, line 938: I'm thinking we should probably create macros or an enum to represent the different tail cail slots, thoughts?

rc-grey (approved):

src/programs.c
RafaelOrtizRC: Yeah, that'll make it less error prone on the `oxidebpf` side.

src/programs.c
RafaelOrtizRC: Turns out this was a bad idea.

PR #7, opened 2021-08-20
Added a helper function for reading struct values

rc-dbogle:
This might also help with stack space usage

### DIFF
--- a/src/programs.c
+++ b/src/programs.c
@@ -1305,9 +1305,23 @@ int kretprobe__ret_ip6_local_out(struct pt_regs *ctx)
     u64 loaded = CRC_LOADED;
     loaded = (u64)bpf_map_lookup_elem(&offsets, &loaded);
 
-    read_value(skbuff_base, CRC_SKBUFF_HEAD, &skb_head, sizeof(skb_head));
-    read_value(skbuff_base, CRC_TRANSPORT_HDR, &transport_header, sizeof(transport_header));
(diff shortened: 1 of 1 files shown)


src/programs.c
```diff
@@ -1293,29 +1305,15 @@ int kretprobe__ret_ip6_local_out(struct pt_regs *ctx)
...
 
-    u64 skb_head_offset = CRC_SKBUFF_HEAD;
-    skb_head_offset = (u64)bpf_map_lookup_elem(&offsets, &skb_head_offset);
-
-    u64 transport_hdr_offset = CRC_TRANSPORT_HDR;
-    transport_hdr_offset = (u64)bpf_map_lookup_elem(&offsets, &transport_hdr_offset);
-
-    u64 network_hdr_offset = CRC_NETWORK_HDR;
-    network_hdr_offset = (u64)bpf_map_lookup_elem(&offsets, &network_hdr_offset);
+    read_value(skbuff_base, CRC_SKBUFF_HEAD, &skb_head, sizeof(skb_head));
```
bjackson41, line 1308: previously we exited early if we were unable to find the offset. do we need to check if the ret is `-1` and exit if so?

```
    if (!skb_head_offset || !transport_hdr_offset || !network_hdr_offset)
    {
        return 0;
    }
```

rc-dbogle: Yea. Good call. I'll update that logic

bjackson41 (approved):

PR #8, opened 2021-08-24
Process events v2

RafaelOrtizRC:
* Adds support for full executable path, filename plus PWD
* Bump ARGV and ENVP grabbing to 4.11+ kernels with `bpf_probe_read_str`


src/programs.c
```diff
@@ -988,26 +1104,99 @@ static __always_inline int enter_exec(syscall_pattern_type_t sp, int fd,
...
     push_telemetry_event(ev);
 
-    bpf_tail_call(ctx, &tail_call_table, 0);
+    bpf_map_update_elem(&telemetry_ids, &pid, &id, BPF_ANY);
 
+    u32 count = 0;
+    char br = 0;
+    //READ_VALUE_N(ev, TE_EXE_PATH, exe, 5); // TODO: is this necessary?
+
+Pwd:;
```
rc-grey, line 1158: Extra semi colon - doesn't really matter in c, but meh

rc-grey (approved):

src/programs.c
RafaelOrtizRC: This is intentional, the label expects a statement.

See: https://stackoverflow.com/questions/8384388/variable-declaration-after-goto-label

src/programs.c
```diff
@@ -988,26 +1104,99 @@ static __always_inline int enter_exec(syscall_pattern_type_t sp, int fd,
...
+    u64 offset = 0;
```
rc-dbogle, line 1159: Seems like this spot has a lot of repetitive code. Could we declare and inline function or macro to clean this up a bit? The `SET_OFFSET` and `FOLLOW_PTR` macros might work I think if you made them a bit more generic.

rc-dbogle (changes requested):
I think there is some repetitive code that we could combine into a function or macro.

src/programs.c
```diff
@@ -330,36 +332,83 @@ static __always_inline syscall_pattern_type_t ptrace_syscall_pattern(u32 request
...
-            bpf_probe_read(&pts, sizeof(pts), ts + *(u32 *)real_parent_offset);                         \
-            if (pts)                                                                                    \
-            {                                                                                           \
-                bpf_probe_read(&ppid, sizeof(ppid), pts + *(u32 *)ppid_offset);                         \
-            }                                                                                           \
-            bpf_probe_read(&luid, sizeof(luid), ts + *(u32 *)luid_offset);                              \
-        }                                                                                               \
+#define FILL_TELEMETRY_SYSCALL_RET(E, SP) \
+
+#define SET_OFFSET(CRC) \
```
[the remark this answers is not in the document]
RafaelOrtizRC: That looks like it should work, I'll try it out. My only concern is inflating the instruction count since the `4_11` probe is very close to the `4096` limit.

src/programs.c
RafaelOrtizRC: ea1cf31

rc-dbogle (approved):
Looks awesome!

PR #12, opened 2021-09-03
fix missing events with legitimate 0

RafaelOrtizRC:
We were missing events with values that were legitimately 0, due to some leftover null checks. I moved the pointer null check into `read_value()` and removed it from the `GET_OFFSET_4_8` macro. Tested and working with loginuid 0 (root).


rc-dbogle (approved):
Looks good.

PR #14, opened 2021-09-17
root namespace pid

RafaelOrtizRC:
We were previously returning the return code from a `fork` family syscall as-is, which gives the child PID in its namespace. This removes several `TE_RETCODE` events for `fork` family calls and replaces them with a single `TE_RETCODE` which contains the child's PID in the root namespace, which we get from `uprobe_copy_process`. This function is called towards the end of `copy_process` and has the child's `task_struct` as its first argument. It appears to be consistently exported in the distros I've tested and is available from 4.11 onwards. `copy_process` and `klp_copy_process` were two other candidates, but were not available on all systems.

### DIFF
--- a/src/programs.c
+++ b/src/programs.c
@@ -2165,8 +2165,12 @@ int kretprobe__ret_sys_clone3(struct pt_regs *ctx)
     return exit_clone3(ctx);
 }
 
-SEC("kprobe/uprobe_copy_process")
-int kprobe__uprobe_copy_process(struct pt_regs *ctx)
+// This probe can generically read the pid from a task_struct at any point
+// where the first argument is a pointer to a task_struct, the event emit
+// is a RETCODE with the correct PID, intended for use with tracing fork,
+// clone, etc.
+SEC("kprobe/read_pid_task_struct")
+int kprobe__read_pid_task_struct(struct pt_regs *ctx)
 {
     // get new current
     void *ts = (void *)PT_REGS_PARM1(ctx);
--- a/src/programs.c
+++ b/src/programs.c
@@ -2197,6 +2197,9 @@ int kprobe__read_pid_task_struct(struct pt_regs *ctx)
     u64 ppid_tgid = (u64) ptgid << 32 | ppid;
     u64 *id = bpf_map_lookup_elem(&telemetry_ids, &ppid_tgid);
 
+    if (!id)
+        return -1;
+
     // send event with ID
     ptelemetry_event_t ev = &(telemetry_event_t){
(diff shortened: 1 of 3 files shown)


src/types.h
```diff
@@ -314,7 +314,11 @@ typedef struct
             char value[VALUE_SIZE];
             char truncated;
         } v;
-        u64 retcode;
+        struct
```
rc-dbogle, line 317: Why put it in a struct instead of just adding another field?

src/programs.c
```diff
@@ -2178,6 +2165,61 @@ int kretprobe__ret_sys_clone3(struct pt_regs *ctx)
...
+        .id = 0,
+        .done = TRUE,
+        .telemetry_type = 0,
+        .u.v = {
+                .value[0] = '\0',
+                .truncated = FALSE,
+                },
+        };
+
+    if (!id)
```
rc-dbogle, line 2211: Seems like this check would make sense to be right after the map lookup. Not strictly required but I think it would flow a bit better that way and then you wouldn't need to initialize the telemetry event in the case that the map lookup failed for some reason

rc-dbogle (changes requested):

src/types.h
RafaelOrtizRC: I'm not crazy about this, but because this change 'replaces' the retcodes for the `clone`/`fork` family calls, I wanted to keep them together somehow. As in, the retcode was _some value_ and but the true `pid` is _some other value_. We probably want a way to grab both the true PID and the namespaced PID.

src/programs.c
RafaelOrtizRC: I stuck it down here because it was the line I was looking at when the verifier complained. It's definitely cleaner up top.
5da8884ffc5ffb81d57a75b463687107cca99979

src/types.h
rc-dbogle: That's fair. If they are mutually exclusive though (we always use one or the other) wouldn't a union make sense? Unless there is a case where we would want both. Either way I don't think it is a huge deal. Just seemed a bit odd at first glance.

RafaelOrtizRC: Even though this should be safe, unions are _discouraged_ by Rust. The use case for a struct would be returning the `fork` retcode (child PID in namespace) alongside the true PID. Although that would require two different probes, with two different messages, so a union is probably more appropriate. 

I think I just talked myself into a union. 567dc34aff74f0aad82406df6b0a206cca61933c

rc-grey (approved):

PR #15, opened 2021-09-23
Dave.bogle/rnd 137/fix udp

rc-dbogle:
Fixing issues with udp outgoing traffic

### DIFF
--- a/src/programs.c
+++ b/src/programs.c
@@ -1418,6 +1418,10 @@ int kretprobe__ret_udp_v6_send_skb(struct pt_regs *ctx)
 
     u64 loaded = CRC_LOADED;
     loaded = (u64)bpf_map_lookup_elem(&offsets, &loaded);
+    if (!loaded)
+    {
+        return 0;
+    }
 
     ret = read_value(skbuff_base, CRC_SKBUFF_HEAD, &skb_head, sizeof(skb_head));
     if (ret == -1)
@@ -1433,11 +1437,6 @@ int kretprobe__ret_udp_v6_send_skb(struct pt_regs *ctx)
 
     ret = read_value(skbuff_base, CRC_NETWORK_HDR, &network_header, sizeof(network_header));
     if (ret == -1)
-    {
-        return 0;
-    }
-
-    if (!loaded)
     {
         return 0;
     }
@@ -1518,6 +1517,10 @@ int kretprobe__ret_ip6_local_out(struct pt_regs *ctx)
 
     u64 loaded = CRC_LOADED;
     loaded = (u64)bpf_map_lookup_elem(&offsets, &loaded);
+    if (!loaded)
+    {
+        return 0;
+    }
 
     ret = read_value(skbuff_base, CRC_SKBUFF_HEAD, &skb_head, sizeof(skb_head));
     if (ret == -1)
@@ -1533,11 +1536,6 @@ int kretprobe__ret_ip6_local_out(struct pt_regs *ctx)
 
     ret = read_value(skbuff_base, CRC_NETWORK_HDR, &network_header, sizeof(network_header));
     if (ret == -1)
-    {
-        return 0;
-    }
-
-    if (!loaded)
     {
         return 0;
     }
@@ -1750,6 +1748,10 @@ int kretprobe__ret_udp_rcv(struct pt_regs *ctx)
 
     u64 loaded = CRC_LOADED;
     loaded = (u64)bpf_map_lookup_elem(&offsets, &loaded);
+    if (!loaded)
+    {
+        return 0;
+    }
 
     ret = read_value(skbuff_base, CRC_SKBUFF_HEAD, &skb_head, sizeof(skb_head));
     if (ret == -1)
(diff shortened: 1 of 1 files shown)


rc-dbogle:
I think a good follow up ticket to this would be to try and make a generic function for processing these network events. I do think there is a bit of code duplication here that could be cleaned up but it would be non trivial

src/programs.c
```diff
@@ -1601,49 +1778,220 @@ int kretprobe__ret___skb_recv_udp(struct pt_regs *ctx)
...
+    ev.telemetry_type = TE_NETWORK;
+    ev.u.network_info.mono_ns = bpf_ktime_get_ns();
+
+    // Get current pid
+    u32 index = (u32)bpf_get_current_pid_tgid();
+
+    // Save the pid in the event structure
+    ev.u.network_info.process.pid = index;
+
+    // Lookup the corresponding *sk that we saved when tcp_v4_connect was called
```
RafaelOrtizRC, line 1833:
```suggestion
    // Lookup the corresponding *sk that we saved when udp_rcv was called
```

src/programs.c
```diff
@@ -1947,6 +2295,13 @@ static __always_inline int exit_clone(struct pt_regs *ctx)
     ev->u.clone_info = clone_info;
     push_telemetry_event(ctx, ev);
 
+    ev->id = *id;
+    bpf_map_delete_elem(&telemetry_ids, &pid_tgid);
+    ev->done = TRUE;
+    ev->telemetry_type = TE_RETCODE;
+    ev->u.retcode = (u32)PT_REGS_RC(ctx);
```
RafaelOrtizRC, line 2302: Same here as the other `retcode`.
```suggestion
    ev->u.r.retcode = (u32)PT_REGS_RC(ctx);
```

src/programs.c
```diff
@@ -2117,60 +2479,6 @@ int kretprobe__ret_sys_clone3(struct pt_regs *ctx)
...
-                .truncated = FALSE,
-                },
-        };
-
-    ev->id = *id;
-    bpf_map_delete_elem(&telemetry_ids, &ppid_tgid);
-    ev->telemetry_type = TE_RETCODE;
-    ev->u.r.pid_tgid = pid_tgid;
-    push_telemetry_event(ctx, ev);
-    return 0;
```
RafaelOrtizRC, line 2171: Please put this back, I'm using it. 😆

RafaelOrtizRC (changes requested):
There's some changes that need to be pulled in from `main`, and I made some suggestions for dedup/cleanup that you can take or ignore.

Looks good! I'd really love to have a dataset on available `/proc/kallsyms` so we don't need to go through this in the first place, though.

src/programs.c
```diff
@@ -1458,8 +1621,8 @@ int kretprobe__ret_inet_csk_accept(struct pt_regs *ctx)
     }
     else if (ev.u.network_info.ip_type == AF_INET6 && loaded)
     {
-        read_value(sk_base, CRC_SOCK_COMMON_DADDR6, &ev.u.network_info.protos.ipv6.dest_addr, sizeof(ev.u.network_info.protos.ipv6.dest_addr));
-        read_value(sk_base, CRC_SOCK_COMMON_SADDR6, &ev.u.network_info.protos.ipv6.src_addr, sizeof(ev.u.network_info.protos.ipv6.src_addr));
+        read_value(sk_base, CRC_SOCK_COMMON_SADDR6, &ev.u.network_info.protos.ipv6.dest_addr, sizeof(ev.u.network_info.protos.ipv6.dest_addr));
+        read_value(sk_base, CRC_SOCK_COMMON_DADDR6, &ev.u.network_info.protos.ipv6.src_addr, sizeof(ev.u.network_info.protos.ipv6.src_addr));
```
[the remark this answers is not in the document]
rc-dbogle: Yea I know it seems weird but that is what gets the data in the right order

src/programs.c
rc-dbogle: Oh that's weird. I don't remember touching that code at all. Not sure how that got deleted

src/programs.c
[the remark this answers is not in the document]
rc-dbogle: I'll add a comment though just to make sure it is clear

src/programs.c
```diff
@@ -1283,13 +1301,57 @@ int kprobe__tcp_v4_connect(struct pt_regs *ctx)
...
 
+SEC("kprobe/udp_send_skb")
+int kprobe__udp_send_skb(struct pt_regs *ctx)
+{
+    _skbuff sk = (_skbuff)PT_REGS_PARM1(ctx);
+    u32 index = (u32)bpf_get_current_pid_tgid();
+
+    bpf_map_update_elem(&udpv4_outgoing_map, &index, &sk, BPF_ANY);
+
+    return 0;
```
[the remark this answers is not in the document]
rc-dbogle: Yup. It worked for a few but unfortunately not all of them. But I like the idea

src/programs.c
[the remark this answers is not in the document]
RafaelOrtizRC: I figured it was something like that, since you flipped them explicitly.

src/programs.c
RafaelOrtizRC: Git probably happily nuked the change in a merge or rebase.

RafaelOrtizRC (approved):
Thanks!

src/programs.c
rc-dbogle: Should be fixed but definitely double check

PR #22, opened 2021-11-02
get more exits

RafaelOrtizRC:
This PR adds two more flexible probes that read inode information and exit code information, respectively.

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -1202,24 +1202,14 @@ static __always_inline int enter_exit(syscall_pattern_type_t sp, int status,
 
     ev->id = id;
     ev->done = FALSE;
-    ev->telemetry_type = TE_EXE_PATH;
-    ev->u.v.truncated = FALSE;
-    long count = 0;
-    count = bpf_probe_read_str(&ev->u.v.value, VALUE_SIZE, (void *)exe);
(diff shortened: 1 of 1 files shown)


src/process-events.c
```diff
@@ -906,6 +906,74 @@ int kretprobe__ret_sys_clone3(struct pt_regs *ctx)
...
+    read_value(ts, CRC_TASK_STRUCT_PID, &pid, sizeof(pid));
+    read_value(ts, CRC_TASK_STRUCT_TGID, &tgid, sizeof(tgid));
+    u64 pid_tgid = (u64)tgid << 32 | pid;
+
+    // Get the inode number and device number
+    u32 i_rdev = 0;
+    u64 i_ino = 0;
+    void *ptr = NULL;
+    read_value(ts, CRC_TASK_STRUCT_MM, &ptr, sizeof(ptr));
+    read_value(ptr, CRC_MM_STRUCT_EXE_FILE, &ptr, sizeof(ptr));
```
rc-dbogle, line 931: One day when we can use BTF this will be so much easier

rc-dbogle (approved):

src/process-events.c
RafaelOrtizRC: Until then, we shall dream.

PR #23, opened 2021-11-16
get exits more reliably and get deeper nested paths

RafaelOrtizRC:
This updates our probes with the following features:
* get deeper nested paths from process starts
* get inode information more reliably
* capture process exits more reliably

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -34,6 +34,15 @@ struct bpf_map_def SEC("maps/read_flush_index") read_flush_index = {
     .namespace = "",
 };
 
+struct bpf_map_def SEC("maps/read_path_skip") read_path_skip = {
+    .type = BPF_MAP_TYPE_PERCPU_ARRAY,
+    .key_size = sizeof(u32),
+    .value_size = sizeof(u64),
+    .max_entries = 1,
+    .pinning = 0,
+    .namespace= "",
+};
+
 struct bpf_map_def SEC("maps/clone_info_store") clone_info_store = {
     .type = BPF_MAP_TYPE_PERCPU_ARRAY,
     .key_size = sizeof(u32),
@@ -243,6 +252,17 @@ static __always_inline void push_telemetry_event(struct pt_regs *ctx, ptelemetry
 
 #define READ_VALUE_N(EV, T, S, N) REPEAT_##N(READ_VALUE(EV, T, S);)
 
+#define SKIP_PATH                                       \
+    if (skipped >= to_skip)                             \
+        goto Send;                                      \
+    /* Skip to the parent directory */                  \
+    bpf_probe_read(&ptr, sizeof(ptr), ptr + parent);    \
+    skipped += 1;                                       \
+    if (!ptr)                                           \
+        goto Send;                                      \
+
+#define SKIP_PATH_N(N) REPEAT_##N(SKIP_PATH;)
+
 #define SEND_PATH                                                           \
     ev->id = id;                                                            \
     ev->done = FALSE;                                                       \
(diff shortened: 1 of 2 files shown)


src/process-events.c
```diff
@@ -906,6 +950,74 @@ int kretprobe__ret_sys_clone3(struct pt_regs *ctx)
...
+    read_value(ptr, CRC_INODE_I_RDEV, &i_rdev, sizeof(i_rdev));
+    read_value(ptr, CRC_INODE_I_INO, &i_ino, sizeof(i_ino));
+
+    u64 *id = bpf_map_lookup_elem(&process_ids, &pid_tgid);
+
+    if (!id)
+        return -1;
+
+    // prepare event
+    ptelemetry_event_t ev = &(telemetry_event_t){
```
bjackson41, line 986: this looks a little awkward to me. are you forcing a pointer for macro use?

src/process-events.c
```diff
@@ -367,10 +397,24 @@ Pwd:;
     SET_OFFSET(CRC_DENTRY_D_PARENT);
     u32 parent = *(u32 *)offset; // offset of d_parent
 
-    SEND_PATH_N(9);
-    bpf_tail_call(ctx, &tail_call_table, SYS_EXECVE_4_8);
+    u64 _to_skip = (u64)bpf_map_lookup_elem(&read_path_skip, &to_skip);
+    if (_to_skip) {
+        __builtin_memcpy(&to_skip, (void*)_to_skip, sizeof(u64));
+        SKIP_PATH_N(100);
```
bjackson41, line 403: whats the reasoning behind these SEND_PATH magic values?

RafaelOrtizRC: This number (I bumped it up to `150`) is determined experimentally, keeping us under `4096` instructions. Sending takes up far more instructions than skipping, so this felt like a good balance.

src/process-events.c
RafaelOrtizRC: Pretty much. To get wide kernel version support, the structure must live on the stack but also we need to provide the helpers with pointers. We don't actually use the regular variable, so it just wastes stack and/or instructions. I use this pattern in a few places.

src/process-events.c
```diff
@@ -906,6 +950,74 @@ int kretprobe__ret_sys_clone3(struct pt_regs *ctx)
...
+    read_value(ts, CRC_TASK_STRUCT_MM, &ptr, sizeof(ptr));
+    read_value(ptr, CRC_MM_STRUCT_EXE_FILE, &ptr, sizeof(ptr));
+    read_value(ptr, CRC_FILE_F_INODE, &ptr, sizeof(ptr));
+    read_value(ptr, CRC_INODE_I_RDEV, &i_rdev, sizeof(i_rdev));
+    read_value(ptr, CRC_INODE_I_INO, &i_ino, sizeof(i_ino));
+
+    u64 *id = bpf_map_lookup_elem(&process_ids, &pid_tgid);
+
+    if (!id)
+        return -1;
```
bjackson41, line 983: IIRC, from the WPM events, we always `return 0`. After reading https://github.com/iovisor/bcc/issues/139, it looks as though returning non-zero has some meaning for kprobes. I wanted to confirm that we are aware of this and returning `-1` was intentional.

RafaelOrtizRC: You're right - we should `return 0` in all cases here and handle partial events on the userspace side.

RafaelOrtizRC: 1c112bb298a8ebf0a28a70535c7755529958190c

bjackson41 (approved):
Looks to be working. Curious how much the `return 0;` made a difference, if any.

PR #24, opened 2021-12-02
Fix dev major minor

rc-dbogle:


src/common.h
```diff
@@ -24,17 +24,30 @@ struct bpf_map_def SEC("maps/offsets") offsets = {
...
+    {
+        bpf_printk("Base should not be NULL\n");
         return -1;
+    }
 
     u64 _offset = (u64)bpf_map_lookup_elem(&offsets, &offset);
     if (_offset)
     {
-        return bpf_probe_read(dest, dest_size, base + *(u32 *)_offset);
+        int ret = bpf_probe_read(dest, dest_size, base + *(u32 *)_offset);
```
bjackson41, line 45: did you mean to commit these changes?

rc-andres: I have removed this and some of other changes that I think slipped through by accident. Let me know if there are other changes you want me to do while Dave is out.

bjackson41: The last thing would be to remove the `printk`s. After that, I think the code changes look good. I'll do some testing to see if any kernel versions have issues with these new offsets.

rc-andres: Just pushed up a change to remove them.

bjackson41 (approved):
Tested on amzn2 and ubuntu. Major/minor was present.

I also tested on centos7-3.10.0-1160.49.1.el7.x86_64 and the major/minor were 0. However, even though this kernel version is outside our range, it looks as though the structs and fields are the same as a 4.20 kernel, so I expected the values to be present.

rc-andres:
> Tested on amzn2 and ubuntu. Major/minor was present.
> 
> I also tested on centos7-3.10.0-1160.49.1.el7.x86_64 and the major/minor were 0. However, even though this kernel version is outside our range, it looks as though the structs and fields are the same as a 4.20 kernel, so I expected the values to be present.

This is definitely strange. Since the sensor currently does not work without correct device major/minor information, I am going to add a check inside the consistency check function in the plugin to verify this is correct and if not default back to audit. I may look into this more later though after we have a better test suite for sensor telemetry.

PR #25, opened 2021-12-07
get the process name during exit_exec* instead of enter_*

rc-andres:
Added the `comm` field to the `file_info` event and removed it from the syscall event. 

This is because:
* We only ever need this for `exec*`. The other types should be able to derive the `comm` from it. 
* During the exec `kprobe` the image has not yet been changed to the new file so the `comm` information is not correct. (i.e., if you run `ls` from `bash` then it would show `bash` as the `comm` of the executed `ls`). To solve this we have moved this to the `kretprobe` instead. 
* The `file_info` event was already an existing event that belongs only to `exec*` and it happens during the `kretprobe` so it made sense to add there instead of making a new event. 

Also renamed `i_rdev` to `i_dev` because rdev is something different that we do not want but it is what we were fetching before.

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -99,8 +99,7 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
     E->u.syscall_info.euid = bpf_get_current_uid_gid() >> 32;        \
     E->u.syscall_info.egid = bpf_get_current_uid_gid() & 0xFFFFFFFF; \
     E->u.syscall_info.mono_ns = bpf_ktime_get_ns();                  \
-    E->u.syscall_info.syscall_pattern = SP;                          \
-    bpf_get_current_comm(E->u.syscall_info.comm, sizeof(E->u.syscall_info.comm));
+    E->u.syscall_info.syscall_pattern = SP;
 
 #define FILL_TELEMETRY_SYSCALL_RET(E, SP)
 
@@ -114,7 +113,7 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
     /* since we're using offsets to read from the structs, we don't need to bother with                 \
      * understanding their structure                                                                    \
      */                                                                                                 \
-    u32 i_rdev = 0;                                                                                     \
+    u32 i_dev = 0;                                                                                      \
     u64 i_ino = 0;                                                                                      \
     void *ts = (void *)bpf_get_current_task();                                                          \
     void *ptr = NULL;                                                                                   \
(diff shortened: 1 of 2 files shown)


src/process-events.c
```diff
@@ -125,7 +124,7 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
         read_value(ptr, CRC_MM_STRUCT_EXE_FILE, &ptr, sizeof(ptr));                                     \
         read_value(ptr, CRC_FILE_F_INODE, &ptr, sizeof(ptr));                                           \
         read_value(ptr, CRC_INODE_I_SB, &sptr, sizeof(sptr));                                           \
-        read_value(sptr, CRC_SBLOCK_S_DEV, &i_rdev, sizeof(i_rdev));                                    \
+        read_value(sptr, CRC_SBLOCK_S_DEV, &i_dev, sizeof(i_dev));                                      \
```
RafaelOrtizRC, line 127: I see what was happening now - this is my fault. I was using the `i_rdev` value which is the device number of the file itself. `i_dev` is the device number of the device that the file is present on. So `i_rdev` will work for char/block devices (not relevant for us), whereas `i_dev` gives the information we're after.

src/process-events.c
```diff
@@ -622,11 +620,13 @@ static __always_inline int exit_exec(struct pt_regs *__ctx, u32 i_rdev, u64 i_in
 
     file_info_t fi = {
         .inode = i_ino,
-        .devmajor = MAJOR(i_rdev),
-        .devminor = MINOR(i_rdev),
-        .value[0] = '\0',
+        .devmajor = MAJOR(i_dev),
+        .devminor = MINOR(i_dev),
+        .comm = {0},
```
RafaelOrtizRC, line 625: Setting `.comm[0] = '\0'` and manually ensuring that it's null-terminated _may_ end up with less instructions than `.comm = {0}`. I think, by default, clang will fill this up with instructions to zero out the field byte-by-byte, which may not be what you want. If the `get_current_comm` helper handles null termination for you, then you may not need to initialize this at all.

RafaelOrtizRC (approved):
Looks good! Just double check that all the kprobes stay under `4096` instructions for backwards-compatibility (I don't see anything that indicates they're not). This should probably be a CI check.

pinealservo (commented):
Looks good as far as I can tell, though going for the more efficient initialization is probably worthwhile given the code size constraints.

src/process-events.c
rc-dbogle: The `bpf_get_current_comm` hasn't always handled NULL termination but it does in 4.11 and on so I think we should be fine

rc-dbogle (commented):
I think lets update the initialization of the comm field as the others had mentioned and then this should be good to go

rc-andres:
> Looks good! Just double check that all the kprobes stay under `4096` instructions for backwards-compatibility (I don't see anything that indicates they're not). This should probably be a CI check.

Just checked and the the kretprobe/ret_sys_exec* sections and the kprobe/read_inode_task_struct section are all still way under 4096. These are the only ones that could have gone up.

PR #26, opened 2021-12-09
Dave.bogle/rnd 217/scripts

rc-dbogle:


src/script-events.c
```diff
@@ -0,0 +1,177 @@
...
+
+    bpf_map_update_elem(&load_script_map, &index, &bprm, BPF_ANY);
+
+    return 0;
+}
+
+SEC("kretprobe/ret_script_load")
+int kretprobe__ret_script_load(struct pt_regs *ctx)
+{
+    char br = 0;
```
RafaelOrtizRC, line 41: I honestly can't remember what was going through my head when I named this variable.

src/script-events.c
```diff
@@ -0,0 +1,177 @@
...
+        bpf_printk("Failed to lookup bprmp\n");
+        return 0;
+    }
+    bpf_map_delete_elem(&load_script_map, &ev.u.script_info.process.pid);
+
+    // Make sure it isn't NULL
+    bprm = (unsigned char *)*bprmp;
+    if (NULL == bprm)
+    {
+        bpf_printk("Failed to deref bprmp\n");
```
RafaelOrtizRC, line 89: Should we remove the `printk`s before merging this in?

src/script-events.c
```diff
@@ -0,0 +1,177 @@
...
+
+    // Output the path
+    bpf_perf_event_output(ctx, &script_events, bpf_get_smp_processor_id(), &ev, sizeof(ev));
+
+    // If the path is / then we don't need to do anything else
+    br = ev.u.script_info.path[0];
+    if (br == '/')
+    {
+        return 0;
+    }
```
RafaelOrtizRC, line 112: Is this checking that we got an absolute path right away? Does the shebang accept a relative path to the PWD?

src/script-events.c
```diff
@@ -0,0 +1,177 @@
...
+    SET_OFFSET(CRC_DENTRY_D_NAME);
+    u32 qstr_len = *(u32 *)offset;
+
+    SET_OFFSET(CRC_QSTR_NAME);
+    u32 name = qstr_len + *(u32 *)offset;
+
+    SET_OFFSET(CRC_DENTRY_D_PARENT);
+    u32 parent = *(u32 *)offset;
+
+#pragma clang loop unroll(full)
```
RafaelOrtizRC, line 142: This passes the verifier all the way back to 4.11? I recall having lots of trouble with older verifiers and loop unrolling.

src/script-events.c
```diff
@@ -0,0 +1,177 @@
...
+            bpf_printk("Failed to read string\n");
+            return 0;
+        }
+
+        br = ev.u.script_info.path[0];
+        if (br == '/')
+        {
+            goto Skip;
+        }
+        bpf_perf_event_output(ctx, &script_events, bpf_get_smp_processor_id(), &ev, sizeof(ev));
```
RafaelOrtizRC, line 164: If we need to account for a relative path and PWD, we should have some logic for situations when the size of a single directory exceeds 128 chars. You should be able to just add on to the offset and continue reading from where you left off, just double check the instruction count stays under 4096.

RafaelOrtizRC (changes requested):
Looks good, just a few questions and one or two suggestions.

src/script-events.c
rc-dbogle: So this is to check how the script was run. If you run something like `ls` or `/usr/bin/ls` then it bypasses all that logic below about getting the current working directory because it already has the absolute path. If you run a script like `./scripts/myscript.sh` then it needs to find the current working directory so that it can pass up the absolute path and read the file contents

src/script-events.c
rc-dbogle: I'll double check just to make sure

src/script-events.c
rc-dbogle: Fair point. I can add that logic in

src/script-events.c
```diff
@@ -35,13 +144,106 @@ int kprobe__script_load(struct pt_regs *ctx)
...
+    ev->done = FALSE;
+    ev->telemetry_type = TE_SCRIPT,
+    ev->u.v.value[0] = '\0';
+    ev->u.v.truncated = FALSE;
+
+    //FILL_TELEMETRY_SYSCALL_EVENT(ev, sp);
+    //ev->u.syscall_info.ppid = ppid;
+    //ev->u.syscall_info.luid = luid;
+
+    // TODO: Do I need this?
```
RafaelOrtizRC, line 171: I don't believe you need this, since this won't happen inside a syscall.

src/script-events.c
```diff
@@ -88,89 +291,51 @@ int kretprobe__ret_script_load(struct pt_regs *ctx)
...
-
-    // Get the offset for pwd in fs_struct
-    SET_OFFSET(CRC_FS_STRUCT_PWD);
-
-    // Read the Dentry pointer for pwd
-    ptr = ptr + *(u32 *)offset; // ptr to pwd
-    if (read_value(ptr, CRC_PATH_DENTRY, &ptr, sizeof(ptr)) < 0)
+    else if (br == '/')
     {
+        bpf_printk("Count < value size. goto Skip\n");
```
RafaelOrtizRC, line 331: `:%s/bpf_printk\(".+"\)\;//gc`

src/script-events.c
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+Send:
+    SEND_PATH_N(10);
+
+    // update to_skip, reuse skipped as index by resetting it
+    skipped = 0;
+    bpf_map_update_elem(&read_path_skip, &skipped, &to_skip, BPF_ANY);
+    // tail call back in
+    bpf_tail_call(ctx, &tail_call_table, 0);
```
RafaelOrtizRC, line 230:
```suggestion
    bpf_tail_call(ctx, &tail_call_table, RET_SCRIPT_LOAD);
```

https://github.com/redcanaryco/redcanary-ebpf-sensor/blob/dave.bogle/RND-217/scripts/src/types.h#L365

src/script-events.c
```diff
@@ -0,0 +1,343 @@
...
+    {
+        return 0;
+    }
+
+    // Just to be safe 0 out the structs
+    telemetry_event_t ev;
+    __builtin_memset(&ev, 0, sizeof(ev));
+
+    // Initialize some of the telemetry event
+    id = bpf_get_prandom_u32();
```
RafaelOrtizRC, line 262: Are you using a kprobe with matching kretprobe here? It looks like that because you check the `&load_script_map` below. In that case, the `id` should probably match across both, unless we're treating the kprobe and kretprobe as separate events.

src/script-events.c
```diff
@@ -0,0 +1,343 @@
...
+
+    // Output the path
+    ev.id = id;
+    ev.done = 0;
+    ev.telemetry_type = TE_CHAR_STR;
+    push_telemetry_event(ctx, &ev);
+    if (count >= sizeof(ev.u.v.value))
+    {
+        bpf_printk("count == to value size. Reading more...\n");
+        READ_CHAR_STR_N(10)
```
RafaelOrtizRC, line 319: Is this trying to get the whole path in the first 10 iterations and, if it can't get it, we go into the tail-call loop in `enter_script_load`?

src/script-events.c
```diff
@@ -0,0 +1,343 @@
...
+    bpf_tail_call(ctx, &tail_call_table, 0);
+
+Skip:
+    skipped = 0;
+    bpf_map_delete_elem(&read_path_skip, &skipped);
+    ev->id = id;
+    return ev;
+}
+
+SEC("kretprobe/ret_script_load")
```
RafaelOrtizRC, line 239: It doesn't look like there's a guard condition in here to skip ahead to the path-sending subroutine. Whenever you tail-call back into this kprobe, it'll re-send all the initial messages.

RafaelOrtizRC (changes requested):
It looks like you have two areas in there where it'll try to loop through the CWD. The tail call flow is also unclear to me.

src/script-events.c
rc-dbogle: Yea sorry. I committed these changes last not forgetting they were going to update this PR

src/script-events.c
rc-dbogle: Yea I left this unimplemented when I committed these changes. I am going to work on this piece today.

src/script-events.c
rc-dbogle: The kprobe doesn't actually emit any data. The kprobe is used to get the linux_bprm pointer and the kretprobe is where most of the work happens since we can check for successful completion of the function at that point. If the function fails we don't want to emit any events or clog up the cache in the plugin

src/script-events.c
RafaelOrtizRC: Oh sorry I thought this was finished. Carry on!

src/script-events.c
RafaelOrtizRC: Oh neat. Good idea!

src/script-events.c
rc-dbogle: Kind of yea. I upped it to 29 in the latest version (4096/144) or (PATH_MAX / VALUE_SIZE). There is a filename filed in the linux_binprm structure. This contains whatever the user used to run the script. If this is an absolute path then all we need to do is read this field and we don't need to call enter_script_load since we don't need the cwd.

src/script-events.c
```diff
@@ -0,0 +1,337 @@
...
+
+    // if the ID already exists, we are tail-calling into ourselves, skip ahead to reading the path
+    u64 p_t = bpf_get_current_pid_tgid();
+    u64 id = (u64)bpf_map_lookup_elem(&process_ids, &p_t);
+    if (id)
+    {
+        __builtin_memcpy(&id, (void *)id, sizeof(u64));
+    }
+
+    // the verifier complains if these are instantiated before the Pwd:; label
```
RafaelOrtizRC, line 163:
```suggestion
```

src/script-events.c
```diff
@@ -0,0 +1,337 @@
...
+    // If we read less than value size and the first character is '/' then we
+    // have an aboslute path and so we can skip getting cwd
+    else if (br == '/')
+    {
+        goto Skip;
+    }
+
+    // If we made it here we need to get the cwd. Add the id we used for our event
+    // the map and then call the handle_pwd function
+    bpf_map_update_elem(&process_ids, &p_t, &id, BPF_ANY);
```
RafaelOrtizRC, line 326: I see what you're doing here. Neat!

src/offsets.h
```diff
@@ -102,3 +102,12 @@ const u64 CRC_SKBUFF_PROTO = 0x42a761c8a2a2b084;
...
 const u64 CRC_SKBUFF_MAC_HDR = 0xd67d00581d6ed4cc;
+
+/* linux_binprm->file */
+const u64 CRC_BPRM_FILE = 0x323e974b140d6eff;
+
+/* linux_binprm->filename */
+const u64 CRC_BPRM_FILENAME = 0xb583ce53935255c5;
+
+/* linux_binprm->interp */
+const u64 CRC_BPRM_INTERP = 0x9f4cedeb0b40dc2c;
```
RafaelOrtizRC, line 113: Nit: the pattern I've been using is `CRC_STRUCT_NAME_MEMBER_NAME`. So the offset of `file` in the `linux_binprm` struct (`/* linux_binprm->file */`) would be `CRC_LINUX_BINPRM_FILE`. This makes it easier to cross reference symbols in Elixir or elsewhere, and ties the constant name directly to the string it represents.

RafaelOrtizRC (approved):
Some minor nits. I haven't tested it, but I assume you've checked on multiple kernel versions? Overall looks alright to me.

src/script-events.c
rc-dbogle: Good catch. Cleaning up comments is important

PR #27, opened 2021-12-09
Partially reverts #23

rc-andres:
This commit partially reverts 619ce922c32ae4ba9913e9b9dbcf9ea3492758a2

We are seeing a bug where all of the pwd in process starts come as
"/". It is fixed when I reverted that commit.

Before:

```
Dec 09 10:34:39.603 INFO proc event: Event {
    pid: 59903,
    parent_id: Some(28658),
    uid: 1000,
    login_uid: Some(1000),
    ktime: 74088176694779,
    kind: Start {
        relative_path: "/bin/ls",
        argv: ["ls", "--color=auto"],
        envp: [],
        cwd: Some("/"), // <~~~ bug
        file_info: FileInfo { inode: 1048662, devmajor: 8, devminor: 1, comm: "ls" }
    }
}
```

After:

```
Dec 09 10:37:08.439 INFO proc event: Event {
    pid: 60575,
    parent_id: Some(28658),
    uid: 1000,
    login_uid: Some(1000),
    ktime: 74236975337488,
    kind: Start {
        relative_path: "/bin/ls",
        argv: ["ls", "--color=auto"],
        envp: [],
        cwd: Some("/home/andres/workspace/spam/"), // <~~~~ correct cwd
        file_info: FileInfo { inode: 1048662, devmajor: 8, devminor: 1, comm: "ls" }
    }
}
```

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -287,7 +287,6 @@ static __always_inline void push_telemetry_event(struct pt_regs *ctx, ptelemetry
         }                                                                   \
         /* we're done here, follow the pointer */                           \
         bpf_probe_read(&ptr, sizeof(ptr), ptr + parent);                    \
-        to_skip += 1;                                                       \
         if (!ptr)                                                           \
             goto Skip;                                                      \
         if (br == '/')                                                      \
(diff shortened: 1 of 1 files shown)


RafaelOrtizRC (changes requested):
Unfortunately, the reason for the original PR was to get us a bigger nested directory depth. It got us up to 150 directories deep when reading the PWD, compared to the 9 we were getting before. If you test it with deeply nested directories you should see the issue pop up. I tested the original PR and it was able to grab nested paths (up to 150 directories deep) without the `/` issue you were seeing. I'm wondering how that issue happened, but in any case we probably don't want to be limited to only 9 directories deep.

rc-andres:
Closed in favor of: https://github.com/redcanaryco/redcanary-ebpf-sensor/pull/28

PR #28, opened 2021-12-09
fix cwd path

rc-andres:
We are currently getting `/` for the path always. This fixes it by correctly handling the skip and tail calls introduced in a prior commit. 

The error was that we were saving a `u32` (the `count` variable) inside a map definition that said that the value is 64 bits. This meant that the upper 32 bits being saved were garbage. When we loaded it as a 64 bit number later and used it to figure out how many segments to skip we were skipping everything since we were getting a giant number. 

This PR fixes it by changing the map definition to a 32 bit value and adding a comment above the `count` variable to note that it needs to stay in-sync with the other variables in the `Pwd` section.


rc-dbogle (approved):
I verified that it fixed the issue we were seeing

PR #29, opened 2021-12-14
Fixing ppid bug

rc-andres:
This fixes two bugs we were seeing in the consistency checker:

1. We were getting a weird fork that had no parent. This was caused by it trying to send an event before CRC_LOADED was written to. 

2. Sometimes the PPID was off by one. This was caused by us using the PID out of the task_struct instead of the TGID. Everywhere else we use the task_struct's TGID as the "pid" and the task struct PID as the "TID". This is also the way the code red sensor expects it. From user space, when someone asks for "what is my current PID", it is actually the TGID what is returned so this matches that expectation as well: https://elixir.bootlin.com/linux/v4.4.160/source/kernel/sys.c#L832

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -148,7 +148,7 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
     if (ts)                                                                                             \
     {                                                                                                   \
         read_value(ts, CRC_TASK_STRUCT_REAL_PARENT, &ptr, sizeof(ptr));                                 \
-        read_value(ptr, CRC_TASK_STRUCT_PID, &ppid, sizeof(ppid));                                      \
+        read_value(ptr, CRC_TASK_STRUCT_TGID, &ppid, sizeof(ppid));                                     \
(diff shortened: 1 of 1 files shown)


src/process-events.c
```diff
@@ -134,15 +134,21 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
      */                                                                                                 \
     u64 offset = CRC_LOADED;                                                                            \
     offset = (u64)bpf_map_lookup_elem(&offsets, &offset); /* squeezing out as much stack as possible */ \
+    /* if CRC_LOADED is not in the map it means we are too early in the ebpf program loading       */   \
+    if (!offset)                                                                                        \
+    {                                                                                                   \
+        bpf_printk("skipping due to missing CRC_LOADED\n");                                             \
```
rc-dbogle, line 140: We shouldn't have any bpf_printk statements
```suggestion
```

rc-dbogle (changes requested):

rc-dbogle (approved):

PR #30, opened 2021-12-27
Updated linux kernel header files to 4.11.

rc-dbogle:
Because of certain eBPF functions that are required our minimum
supported version is 4.11.
Updated build toolchain to v8 instead of v6

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM ubuntu:16.04
 
 RUN apt-get update
-RUN apt-get install -y llvm-6.0 clang-6.0 libclang-6.0-dev \
-    linux-headers-4.4.0-98-generic linux-headers-4.10.0-14-generic \
+RUN apt-get install -y llvm-8 clang-8 libclang-8-dev \
+    linux-headers-4.11.0-14-generic linux-headers-4.10.0-14-generic \
     make binutils curl coreutils
 
 WORKDIR /src
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,10 @@ SRC = src
 SRCS = $(wildcard $(SRC)/*.c)
 OBJS = $(patsubst $(SRC)/%.c,$(SRC)/%,$(SRCS))
 OBJS_WRAPPED = $(OBJS)
-CC = clang-6.0
-LLC = llc-6.0
-OPT = opt-6.0
-LLVM_DIS = llvm-dis-6.0
+CC = clang-8
+LLC = llc-8
+OPT = opt-8
+LLVM_DIS = llvm-dis-8
 CFLAGS += \
 	-D__KERNEL__ \
 	-D__BPF_TRACING__ \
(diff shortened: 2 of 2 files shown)


rc-dbogle:
By upping it from the 4.4 headers to the 4.11 headers we will have access to more features in BPF that we didn't have access to before such as newer map types and helper functions

Makefile
```diff
@@ -36,7 +36,7 @@ TARGET = -target aarch64
 else ifeq ($(ARCH),x86_64)
 CFLAGS += -D__ASM_SYSREG_H
 KERNEL_ARCH_NAME = x86
-KERNEL_HEADER_VERSION ?= 4.4.0-98-generic
+KERNEL_HEADER_VERSION ?= 4.11.0-14-generic
```
rc-andres, line 39: should we bump the aarch64 one also? It's currently at 4.10.0-14

Makefile
```diff
@@ -76,8 +76,8 @@ endif
 
 depends:
 	apt-get update
-	apt-get install -y llvm-6.0 clang-6.0 libclang-6.0-dev \
-		linux-headers-4.4.0-98-generic linux-headers-4.10.0-14-generic \
+	apt-get install -y llvm-8 clang-8 libclang-8-dev \
+		linux-headers-4.11.0-14-generic linux-headers-4.10.0-14-generic \
```
rc-andres, line 80: why not `linux-headers-${KERNEL_HEADER_VERSION}`? Do we need both the aarch64 and x64_64 header versions on every install? I would assume we only want the one we are currently building for. 

Similarly, maybe not super important but if the llvm and clang versions are so co-dependent maybe we should have a variable for the version and do something like:

```bash

CLANG_VERSION=8

CC = clang-${CLANG_VERSION}
LLC = llc-${CLANG_VERSION}
OPT = opt-${CLANG_VERSION}
LLVM_DIS = llvm-dis-${CLANG_VERSION}

apt-get install -y llvm-${CLANG_VERSION} clang-${CLANG_VERSION} libclang-${CLANG_VERSION}-dev \
        linux-headers-${KERNEL_HEADER_VERSION}

```

Although that may look too messy unsure.

Makefile
rc-dbogle: Honestly I don't know much about the ARM build. @bjackson41 do you have any insights on this?

bjackson41: 4.10.0-14 was a minimum to get things working on aarch64. I can do some testing to see if this 4.11.0-14 can work as well.

Makefile
bjackson41: This looks cleaner than what we currently have.

rc-dbogle: Yea I think that should work fine

Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM ubuntu:16.04
 
 RUN apt-get update
-RUN apt-get install -y llvm-6.0 clang-6.0 libclang-6.0-dev \
-    linux-headers-4.4.0-98-generic linux-headers-4.10.0-14-generic \
+RUN apt-get install -y llvm-8 clang-8 libclang-8-dev \
```
rc-dbogle, line 4: I noticed that the dockerfile does basically the same command as the Makefile? Is this necessary or should we just let it happen in the Makefile? Seems like it will just do the same thing twice

Makefile
rc-dbogle: There were some build issues on ARM when trying to upgrade to 4.11. I made a ticket to track it but I think it might be something that should be addressed in a separate PR

rc-andres (approved):

PR #31, opened 2022-01-06
Filter out failed execs

rc-andres:
Do not emit events for exec* syscalls with a non-zero retcode as no process was started. 

I have also removed the unused and unneeded `done` field.

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -612,6 +612,10 @@ static __always_inline int exit_exec(struct pt_regs *__ctx, u32 i_dev, u64 i_ino
 
     bpf_map_delete_elem(&process_ids, &pid_tgid);
 
+    u32 retcode = (u32)PT_REGS_RC(__ctx);
+    if (retcode != 0) {
+        goto Flush;
+    }
 
     ptelemetry_event_t ev = &(telemetry_event_t){
         .id = *id,
@@ -629,7 +633,7 @@ static __always_inline int exit_exec(struct pt_regs *__ctx, u32 i_dev, u64 i_ino
     // re-use the same ev to save stack space
     ev->id = *id;
     ev->telemetry_type = TE_RETCODE;
-    ev->u.r.retcode = (u32)PT_REGS_RC(__ctx);
+    ev->u.r.retcode = retcode;
     push_telemetry_event(__ctx, ev);
 
 Flush:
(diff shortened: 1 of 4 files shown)


src/process-events.c
```diff
@@ -616,34 +610,30 @@ static __always_inline int exit_exec(struct pt_regs *__ctx, u32 i_dev, u64 i_ino
...
 
-    file_info_t fi = {
-        .inode = i_ino,
-        .devmajor = MAJOR(i_dev),
-        .devminor = MINOR(i_dev),
-    };
+    u32 retcode = (u32)PT_REGS_RC(__ctx);
+    if (retcode != 0) {
+        goto Flush;
+    }
```
RafaelOrtizRC, line 618: Tested with the below program and can confirm that an event is emit for `ls` but not for `lol`.

```
#include <unistd.h>

void main() {
	char *arg[] = { "/no/file/lol", NULL };
	char *env[] = { NULL };
	execve("/no/file/lol", arg, env);
	execve("/usr/bin/ls", arg, env);
}
```

src/types.h
```diff
@@ -319,7 +319,6 @@ typedef struct
 typedef struct
 {
     u64 id;
-    u32 done;
```
RafaelOrtizRC, line 322: Thanks for cleaning this up while you were here!

RafaelOrtizRC (approved):
The logic here looks sound and is functioning on Ubuntu 20.04. I'm having trouble with other distros though.

rc-andres:
> The logic here looks sound and is functioning on Ubuntu 20.04. I'm having trouble with other distros though.

I responded in the PR in our private repo but essentially I believe the errors in the other distros are unrelated. Some of those distros (namely Centos 7) don't seem to support the program sizes we need for `sys_execve_4_11` which we need to get the correct binary path. Other ones it's from some distros/kernel version pairs missing some of offset files. 

I tried it Centos8.5 and it works with kernel:
```
[andres@localhost cwp-plugins]$ uname -r
4.18.0-348.7.1.el8_5.x86_64
```

rc-dbogle:
@RafaelOrtizRC I have tried it on Centos 8, Oracle 8, Fedora 33 and Ubuntu and it all passed the consistency check

rc-dbogle (approved):
Tried it on a few other platforms and it seems to function fine

rc-dbogle:
We did see a few other issues on some older platforms but those should be addressed separately

RafaelOrtizRC:
@rc-dbogle 

> We did see a few other issues on some older platforms but those should be addressed separately

Are these the issues I had on older Ubuntu distros and 3.x EL kernels?

PR #32, opened 2022-01-19
Fix perf buffer and discard bad events

rc-andres:
This contains a couple of fixes:

* Set `max_entries` to 0 on perfmap so oxidebpf can default them appropiately

* Use `BPF_F_CURRENT_CPU` instead of calling for `bpf_get_smp_processor_id()` as per the docs: https://man7.org/linux/man-pages/man7/bpf-helpers.7.html.

> flags can be set to BPF_F_CURRENT_CPU to indicate that the index of the current CPU core should be used.
  
* Send a discard event forward when we process failed exec starts. This will let the Rust code pop that telemetry id out of its LRU cache.

### DIFF
--- a/src/cred-events.c
+++ b/src/cred-events.c
@@ -62,7 +62,7 @@ static __always_inline int dispatch_credentials_event(struct pt_regs *__ctx)
     {
         bpf_perf_event_output(__ctx,
                               &cred_events,
-                              bpf_get_smp_processor_id(),
+                              BPF_F_CURRENT_CPU,
                               pcreds,
                               sizeof(*pcreds));
     }
--- a/src/cred-events.c
+++ b/src/cred-events.c
@@ -18,7 +18,7 @@ struct bpf_map_def SEC("maps/cred_events") cred_events = {
     .type = BPF_MAP_TYPE_PERF_EVENT_ARRAY,
     .key_size = sizeof(u32),
     .value_size = sizeof(u32),
-    .max_entries = 1024,
+    .max_entries = 0, // let oxidebpf set it to num_cpus
     .pinning = 0,
     .namespace = "",
 };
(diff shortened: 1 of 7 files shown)


rc-dbogle (approved):
Looks good to me!

src/cred-events.c
```diff
@@ -62,7 +62,7 @@ static __always_inline int dispatch_credentials_event(struct pt_regs *__ctx)
     {
         bpf_perf_event_output(__ctx,
                               &cred_events,
-                              bpf_get_smp_processor_id(),
+                              BPF_F_CURRENT_CPU,
```
RafaelOrtizRC, line 65: I want to say this was a holdover from when we wanted to support kernels as old as 4.4. The flag is fewer instructions, so this is a nice win.

src/cred-events.c
```diff
@@ -18,7 +18,7 @@ struct bpf_map_def SEC("maps/cred_events") cred_events = {
     .type = BPF_MAP_TYPE_PERF_EVENT_ARRAY,
     .key_size = sizeof(u32),
     .value_size = sizeof(u32),
-    .max_entries = 1024,
+    .max_entries = 0, // let oxidebpf set it to num_cpus
```
RafaelOrtizRC, line 21: Excellent!

RafaelOrtizRC (approved):
I don't think it's anywhere close, but since you added insns it's always good to double check that we're not exceeding 4096 instructions anywhere. Overall, looks good to me.

rc-andres:
> I don't think it's anywhere close, but since you added insns it's always good to double check that we're not exceeding 4096 instructions anywhere. Overall, looks good to me.

Weirdly enough it is not (I think we probably passed it at another point) and yet it still works on my Centos 8 kernel 4.18.x. 

Just to be on the safe side though I have reduced the SKIP_PATH_N to 125 to go back under 4k

PR #33, opened 2022-01-22
set the tgid as the pid on network info

rc-andres:
the kernel pid is what the userspace calls tid so we want the tgid instead. 

This was causing code red to belief that some events were coming out of order because it couldn't find the PID in its list of running events but in reality it was because we were using the TID instead of the PID. 

https://redcanary.atlassian.net/browse/CWP-426


rc-dbogle (approved):
Good catch!

PR #34, opened 2022-01-25
Fix execveat args

rc-andres:
https://redcanary.atlassian.net/browse/CWP-432

telemetry for `execveat` was returning completely bogus data for argv. This was because the tail call that we were using was specific to the argument order of `execve` (filename, argv, envp,...). This PR adds a different tail calls specific to `execveat` and renames the tail calls to be more specific. 

I also made the following small changes:

* Changed the `ptr` to be an actual `char *` rather than just `u64`. This should in theory be safer across different 32bit vs 64bit systems. 
* Check the the return code from `bpf_probe_read_str` is not a negative number. This was causing us to read invalid memory again and again rather than exiting on the first error. We are no longer reading invalid memory since we fixed the core issue but this should make it more fail safe for the future. Also the ret code before was being stored in a long, but the kernel code actually returns an int. The docs do say `long` but that is actually incorrect and it makes it seem like we get back a HUGE string size in the error case. 
* Merged `READ_VALUE_N` and `__READ_LOOP_N` as they were basically identical. 

Note that to test this the plugins side needs to also update to the new tail code slots.


src/process-events.c
```diff
@@ -444,12 +447,10 @@ int BPF_KPROBE_SYSCALL(kprobe__sys_exec_tc_argv,
...
     }
 
-    long count = 0;
-    u32 ii = 0;
-    // explicit copy from pointer to stack, satisfy verifier
-    __builtin_memcpy(&ii, pii, sizeof(u32));
-    u64 ptr = 0;
-    u64 ret = 0;
+    int count = 0;
+    u32 ii = pii[0];
```
pinealservo: Not sure why you're doing `pii[0]` instead of `*pii` here?  They should do the same thing, but using the array syntax suggests that there's an array involved.

rc-andres: these are stored in a single element array specified at the start of the file:

```c
struct bpf_map_def SEC("maps/read_flush_index") read_flush_index = {
    .type = BPF_MAP_TYPE_PERCPU_ARRAY,
    .key_size = sizeof(u32),
    .value_size = sizeof(u32),
    .max_entries = 1,
    .pinning = 0,
    .namespace = "",
};
```

This is an array per cpu, where each array holds at most 1 element of key and value sized to a `u32`. 

I figured this would make it more obvious that an array is somehow involved as you said, but it should be equivalent.

pinealservo: Ah, ok, but I think `bpf_map_lookup_elem` is already doing the array indexing and is just returning a pointer to the single indexed element.  This runaround is apparently so the lookup function is generic across different map types; e.g. one of the hashmap-based maps wouldn't be just a straightforward array index.

In other words, the pointer returned is conceptually some single element from somewhere in a map, and not the first of an array of values, even if it actually is in this case the first of a single-element array.

rc-andres: Ah you are right! I will make that be just a dereference then!

pinealservo:
So, I am unsure about the "right" thing to do with regard to the change from two `u64` values to a `char *` and `int`.  It seems that the `int` is a workaround for a bug in some versions of the `bpf_probe_read` function in the kernel which truncate a `long` result from the lower-level implementation to an `int` which makes 32-bit error codes get unsigned expansions to `u64` in the eBPF call interface.  The newest version of the kernel I looked at (5.12.19) doesn't seem to have this problem, so casting the `u64` to a `long` as the docs suggest would give you negative error codes again, but we probably don't have the luxury of only targeting these kernels.

While using a 64-bit value is more "correct" in some sense, we'll never use enough of the range for the difference to matter and just casting it to int makes it much harder to accidentally misinterpret an error value. The 64-to-32 bit truncation will just remove a bunch of leading 1s or 0s depending on the sign of the number for any transfer size/error code that is meaningful to us.

I don't have strong opinions either way about storing the pointer in a `char *` vs `u64`, as the code path is going to cast it from one to the other at some point either way.

So, my vote is to leave it as Andres changed it here, though I'd welcome any other viewpoints.

rc-dbogle (approved):
Thanks for the extensive testing to find and address this issue

PR #35, opened 2022-01-28
Fix execveat path

rc-andres:
https://redcanary.atlassian.net/browse/CWP-433

`execveat` is kind of a funky syscall in that it splits the finding of the executable into the `fd` and the `filename` and has certain options to ignore one or the other. This was making the executable path that we send to `code-red` incorrect and/or empty as we were only looking at the `filename` since it was basically a copy of what `execve` does. 

In order to get the correct path without re-creating the logic in the kernel this PR hooks into the return value of `do_open_execat` which returns `file *`. I have verified that this function exists in both kernel 4.11 and the latest kernel. This function is called in shared common between `execve` and `execveat` to get the "file" to be executed. Since this code is shared between these two syscalls, the ebpf program makes sure the path sending logic only happens during `execveat`. This is because we would like `execve` to still have the correct idea of what is PWD vs relative path, while in `execveat` we still send the PWD but in reality the "relative path" to the executable will always be an absolute path.

The existing code that goes through a `path` "recursively" for PWD is re-used to send the full executable path for execveat. To do this a new type of telemetry event was created since the existing execfilename event expects it in forward order while the path per segment logic sends it in reverse (from file to root) order. 

A lot of the code in here was refactoring while I was trying to understand how it all worked and also making the path sending code re-usable so I didn't have to copy paste it. I have verified that we are not exceeding our 4096 instruction limit in any of the ebpf programs. I have also removed some types and events that were un used to free up some instruction and stack space. 

To test you can use the cwp-plugins branch at: https://github.com/redcanaryco/cwp-plugins/pull/97

### Both `fd` and `path`

```
Start {
    relative_path: "/usr/bin/id",
    argv: ["first_arg", "second arg", "12345"],
    cwd: Some("/home/andres/workspace/playground/"),
    file_info: FileInfo { inode: 134678, devmajor: 8, devminor: 1, comm: "id" }
}
```

### Only `fd` using AT_EMPTY_PATH

```
Start {
    relative_path: "/usr/bin/id",
    argv: ["first_arg", "second arg", "12345"],
    cwd: Some("/home/andres/workspace/playground/"),
    file_info: FileInfo { inode: 134678, devmajor: 8, devminor: 1, comm: "3" }
 }
 ```
 
 The relative path shows the absolute path pointed by the fd. The `comm` is `3` because that is the number of the `fd`. This comes from the kernel (`get_current_comm`) and it's the same thing we would get in audit. 

### Only `fd` but with a `memfd` instead of a file

```
Start {
    relative_path: "memfd:not-suspicious",
    argv: ["first_arg", "second arg", "12345"],
    cwd: Some("/home/andres/workspace/playground/"),
    file_info: FileInfo { inode: 1533313, devmajor: 0, devminor: 1, comm: "3" }
}
```

The name of the `memfd` is `non-suspicious` so we know that the "executable name" will include the fact that it was run from a `memfd` which is important. 

### Only `path` (using `AT_FDCWD` as `fd` and a relative path)

```
Start {
    relative_path: "/usr/bin/id",
    argv: ["first_arg", "second arg", "12345"],
    cwd: Some("/home/andres/workspace/playground/"),
    file_info: FileInfo { inode: 134678, devmajor: 8, devminor: 1, comm: "id" }
}
```

Note that to run this one the path had to be specified relative (i.e., `../../../../usr/bin/id`) but currently for all `execveat` we just get the full absolute path. I could have special handling logic if the `fd` is `AT_FDCWD` if we think it's worth keeping the knowledge that the path was run as a relative path instead of as an absolute path. Either way the path being given is correct.

### CODE REVIEW

File: src/process-events.c [hunk 1]
```diff
@@ -148,15 +148,6 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
...
         read_value(ts, CRC_TASK_STRUCT_LOGINUID, &luid, sizeof(luid));                                  \
-        read_value(ts, CRC_TASK_STRUCT_MM, &ptr, sizeof(ptr));                                          \
-        read_value(ptr, CRC_MM_STRUCT_EXE_FILE, &ptr, sizeof(ptr));                                     \
-        SET_OFFSET(CRC_FILE_F_PATH);                                                                    \
-        ptr = ptr + *(u32 *)offset; /* ptr to f_path */                                                 \
-        read_value(ptr, CRC_PATH_DENTRY, &ptr, sizeof(ptr));                                            \
-        SET_OFFSET(CRC_DENTRY_D_NAME);                                                                  \
-        ptr = ptr + *(u32 *)offset; /* ptr to d_name */                                                 \
-        read_value(ptr, CRC_QSTR_LEN, &length, sizeof(length));                                         \
-        read_value(ptr, CRC_QSTR_NAME, &exe, sizeof(exe));                                              \
```
pinealservo, line 159: Hmm, it looks like this ought to retrieve the name associated with the file that's retrieved via `/proc/[pid]/exe`.  I wonder why we weren't using this?

rc-andres: I think there's a couple of gotchas but maybe we should:

1. We want the entire path to the executable. This used to get just the name.
2. We were doing this for *every* syscall which is a waste, only execve* really needs to know the exec path
3. At the time call for this we are still in the executable prior to the syscall finishing. In the case of execve* (the ones were we actually need this information) the exe name here would be for the executable we are replacing not the one we are launching. 

All those gotchas can be addressed though I think:

1. We can use the function I pulled out in the refactor to trace the `path` back step by step to get the full path.
2. We can have it just be used for the `execveat` syscall. The reason to not do it for `execve` is so we can still tell if a executable was run using a relative or absolute path
3. We can do this in the `kretprobe` so the `execveat` has finalized and the image has been replaced. 

I think this may actually be better than adding that new program that hooks into that random function. Now I feel silly not thinking of doing that when I was deleting this code. I can give it a shot.

pinealservo: So, I didn't see in the diffs where we got the other file identity pieces (inode, dev major and minor), but they're also available via the `task->mm->exe_file->f_path->dentry` route. Apparently the `f_path` value also has a pointer to the `vfsmount` structure, which might have some fields that could help us find the path in the filesystem, though I'm not entirely sure about how that would work.

Anyway, I didn't see anything that jumped out at me as concerning, and I really liked the examples you set up in the PR description.

pinealservo (approved):
We might eventually want to get some of the unused fields back, but it seems OK with me to remove them for now given the space constraints.

rc-dbogle (approved):

src/process-events.c
```diff
@@ -263,22 +258,136 @@ static __always_inline void push_telemetry_event(struct pt_regs *ctx, ptelemetry
...
+
+    void *ptr = NULL;
+    read_value(ts, CRC_TASK_STRUCT_MM, &ptr, sizeof(ptr));
+    read_value(ptr, CRC_MM_STRUCT_EXE_FILE, &ptr, sizeof(ptr));
+
+    return ptr;
+}
+
+// it's argument should be a pointer to a file
+static __always_inline file_info_t extract_file_info(void *ptr)
```
rc-dbogle, line 312: Ditto

src/process-events.c
```diff
@@ -263,22 +258,136 @@ static __always_inline void push_telemetry_event(struct pt_regs *ctx, ptelemetry
...
+        ev->u.r.retcode = retcode;
+        push_telemetry_event(ctx, ev);
+
+        return true;
+    }
+
+    return false;
+}
+
+static __always_inline void* get_current_exe()
```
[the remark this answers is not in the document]
rc-andres: As of right now no one else is calling for `bpf_get_current_task` so I rather not extract it outside this file until it is actually common across different telemetry sources.

src/process-events.c
rc-andres: As of right now no one else is calling for `bpf_get_current_task` so I rather not extract it outside this file until it is actually common across different telemetry sources.

PR #38, opened 2022-02-19
feat: filter out kernel processes

rc-vbundage:
[Jira Story - CWP-444](https://redcanary.atlassian.net/browse/CWP-444?atlOrigin=eyJpIjoiMjZlMjIzMWNhZTA1NDM0M2EyMzFmYjZkYjMxMDI0OGYiLCJwIjoiaiJ9)

At the moment the eBPF sensor collects telemetry on process exits from Kernel processes. This PR would filter this type of telemetry out to reduce the noise in the output. 

I tested this by greping for process exits with a `uid` of `0` and `login_uid` of `None` when running `proc_events` while also running the audit sensor. Stopping execution of `cfsvcd` would cause a lot of these process exit events to show in the eBPF telemetry. 

Before applying filter:
Screen Shot 2022-02-18 at 5 39 48 PM

After applying filter:
Screen Shot 2022-02-18 at 5 45 55 PM

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -117,14 +117,14 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
      */                                                                                                 \
     void *ts = (void *)bpf_get_current_task();                                                          \
     void *ptr = NULL;                                                                                   \
-    /* check mm field of task_struct                                                                    \
-     * skip kernel processes as they have an mm field of NULL                                           \
-     */                                                                                                 \
     void *mmptr = NULL;                                                                                 \
-    read_value(ts, CRC_TASK_STRUCT_MM, &mmptr, sizeof(mmptr));                                          \
-    if (!mmptr) return 0;                                                                               \
     if (ts)                                                                                             \
     {                                                                                                   \
+        /* check mm field of task_struct                                                                \
+         * skip kernel processes as they have an mm field of NULL                                       \
+         */                                                                                             \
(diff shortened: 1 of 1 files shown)


src/process-events.c
```diff
@@ -146,6 +146,13 @@ int kprobe__do_mount(struct pt_regs *ctx)
 
 static __always_inline void push_telemetry_event(struct pt_regs *ctx, ptelemetry_event_t ev)
 {
+    // check mm field of task_struct
+    // skip pushing kernel process events as they have an mm field of NULL
+    void *ts = (void *)bpf_get_current_task();
+    void *ptr = NULL;
+    read_value(ts, CRC_TASK_STRUCT_MM, &ptr, sizeof(ptr));
+    if (!ptr) return;
```
rc-vbundage, line 154: I'm not sure if this is the appropriate place to implement this filtering. My reasoning was that it would be better here than duplicating these lines in several places.

rc-andres: I think that having it here is a bit of an overkill. Doing this check on every push is probably not necessary since some syscalls send multiple events but really we just need to check it once per syscall (or maybe even only on exit, unsure if the other syscalls would be triggered at all by kernel processes). 

It might fit better either explicitly in the exit syscalls for kernels > 4.8, or in `GET_OFFSETS_4_8` since we already get the current task there.

rc-andres (approved):

PR #40, opened 2022-04-14
Use #pragma unroll rather than macros for loops

rc-andres:
I started looking at this code again to address a race condition (not fixed in this PR) and got annoyed by how hard some of it is to read. In service of making changes easier I have replaced the `#REPEAT` macro in `process_events.c` with the `clang` `#pragma unroll` directive. This directive does the loop unrolling for us so we don't need fancy macros to simulate for-loops anymore. This still has the limitation that they have to be bounded but that's good enough for us. 

I also went ahead and added a version of push event that does not clear the event. This was pretty expensive (# of instructions) in our loops especially since in those loops we want to keep the same `id` and `telemetry_type`, and just change the data. I still kept the old one and only changed it for the loop code to avoid any risk on some areas where we are depending on that variable being cleared. 

All in all these had a good effect in our number of instructions. 

Before the change:
`sys_execveat_411`: 3565
`sys_execve`: 3836
`sys_execve_tc_argv` AND `sys_execveat_tc_argv`  : 1701

After the change:
`sys_execveat_411`: 2863
`sys_execve`: 2987
`sys_execve_tc_argv` AND `sys_execveat_tc_argv`  : 878

This should give us room to add more code to these programs which we'll need to do in an upcoming PR to fix the aforementioned race condition. 

I have tested that the consistency check still passes and that it still works with a high number of long-named arguments and a deeply nested long-named folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # redcanary-ebpf-sensor
 
 This project consists of a variety of eBPF applications aimed at gathering events
-of interest for [Red Canary's Cloud Workload Protection](https://redcanary.com/products/cloud-workload-protection/) 
+of interest for [Red Canary's Cloud Workload Protection](https://redcanary.com/products/cloud-workload-protection/)
 product.
 
-These applications do not use [BCC](https://github.com/iovisor/bcc) to build. The 
+These applications do not use [BCC](https://github.com/iovisor/bcc) to build. The
 main objective of this design is to have a compile once, run everywhere application.
 
 To build this project run
@@ -12,3 +12,21 @@ To build this project run
 
 A vscode cpp properties files has been included. Make sure to update the include path with the path
 on your local system where the kernel header files are located
+
+## Validate Instruction Count
+
+Due to older kernel limitations (< 5.2) the instruction limit for our
+ebpf programs is 4096. This was changed in Kernel 5.2+ to be 1 million
+but we cannot rely on that at this time. To verify that we aren't
+going over the limit, after modifying an ebpf program run it through
+`llvm-objdump` and check its instruction count:
+
+```bash
+llvm-objdump -d <PATH_TO_COMPILED_FILE> -j <SPECIFIC_SECTION_TO_ANALYZE> | less
+```
+
+You may ommit the `-j <SPECIFIC_SECTION_TO_ANALYZE>` if you want to
+check all the sections at the same time.
+
+eBPF programs can branch (but not jump back!) so make sure to check
+that none of the branches go over the 4096 instructions limit.
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -163,59 +163,6 @@ static __always_inline void push_telemetry_event_reuse(struct pt_regs *ctx, ptel
     bpf_perf_event_output(ctx, &process_events, BPF_F_CURRENT_CPU, ev, sizeof(*ev));
 }
 
-#define SKIP_PATH                                    \
-    if (skipped >= to_skip)                          \
-        goto Send;                                   \
-    /* Skip to the parent directory */               \
-    bpf_probe_read(&ptr, sizeof(ptr), ptr + parent); \
-    skipped += 1;                                    \
-    if (!ptr)                                        \
-        goto Send;
-
-#define SKIP_PATH_N(N) REPEAT_##N(SKIP_PATH;)
-
-#define SEND_PATH(T)                                                        \
-    if (br == 0)                                                            \
-    {                                                                       \
-        if (bpf_probe_read(&offset, sizeof(offset), ptr + name) < 0)        \
-            goto Skip;                                                      \
-        if (!offset)                                                        \
-            goto Skip;                                                      \
-    }                                                                       \
-    ev->telemetry_type = T;                                                 \
-    ev->id = id;                                                            \
-    __builtin_memset(&ev->u.v.value, 0, VALUE_SIZE);                        \
-    count = bpf_probe_read_str(&ev->u.v.value, VALUE_SIZE, offset);         \
-    if (count < 0)                                                          \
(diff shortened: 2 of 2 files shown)


pinealservo (approved):
Looks OK to me; better be careful with re-use of structures that you don't leave bits set that will change things unintentionally.  I'm pretty sure I had some bugs in either audit or openbsm parsing due to something like that, though I don't remember the details anymore.

RafaelOrtizRC:
I mentioned it internally but I'll put it here for the open source record. We didn't originally create `#REPEAT` because we're masochists (maybe a little, this is eBPF) but because `#pragma unroll` resulted in code that was shorter, but too complex for certain older verifiers. The verbose but simpler instructions generated by `#REPEAT` passed verification more easily. This may no longer be the case, but it's definitely worth testing in as many places as we can.

rc-andres:
> I mentioned it internally but I'll put it here for the open source record. We didn't originally create `#REPEAT` because we're masochists (maybe a little, this is eBPF) but because `#pragma unroll` resulted in code that was shorter, but too complex for certain older verifiers. The verbose but simpler instructions generated by `#REPEAT` passed verification more easily. This may no longer be the case, but it's definitely worth testing in as many places as we can.

I have tested this on:
Ubuntu 18.04 w/ 4.15.0-20
Ubuntu 18.04 w/ 5.4.0-107
CentOS with 3.10.0-1160

All without any problems. I think as you mentioned in slack this may have been a relic of when we were trying to support kernels 4.8 or 4.4. This is no longer the case since we use `BPF_FUNC_probe_read_str ` heavily and that was introduced in 4.11. I didn't do a test on 4.11 itself since we don't have offsets for 4.11 on Ubuntu 18.04. 

I am going to try to do a test on Ubuntu 16.04 with a kernel that should match our minimum version better but assuming that also passes I am going to consider this safe.

README.md
```diff
@@ -1,14 +1,32 @@
...
 
 To build this project run
 `docker-compose run  --rm ebpf make all`
 
 A vscode cpp properties files has been included. Make sure to update the include path with the path
-on your local system where the kernel header files are located
\ No newline at end of file
+on your local system where the kernel header files are located
+
+## Validate Instruction Count
```
RafaelOrtizRC, line 16: Great idea adding this to the readme

src/process-events.c
```diff
@@ -156,104 +156,12 @@ static __always_inline void push_telemetry_event(struct pt_regs *ctx, ptelemetry
...
-            off = off + VALUE_SIZE;                                                    \
-        }                                                                              \
-        else                                                                           \
-        {                                                                              \
-            br = 1;                                                                    \
-        }                                                                              \
-        push_telemetry_event(ctx, EV);                                                 \
-    }
-
-#define READ_VALUE_N(EV, T, PTR, N) REPEAT_##N(READ_VALUE(EV, T, PTR);)
```
RafaelOrtizRC, line 182: RIP

src/process-events.c
```diff
@@ -374,40 +349,48 @@ static __always_inline int process_argv(struct pt_regs *ctx,
...
+        arg_num = *parg_num;
+    }
 
     // this number was arrived at experimentally, increasing it will result in too many
     // instructions for older kernels
-    READ_LOOP_N(argv, TE_COMMAND_LINE, 6);
-
-    bpf_map_update_elem(&read_flush_index, &index, &ii, BPF_ANY);
+    #pragma unroll
+    for (int i = 0; i < 6; i++) {
```
RafaelOrtizRC, line 375: Not for this PR, but it may be worth going back and seeing if we can raise some of these limits since `#pragma unroll` should generate fewer instructions.

src/process-events.c
```diff
@@ -502,17 +485,18 @@ int BPF_KPROBE_SYSCALL(kprobe__sys_execve_4_11,
 
     // create the event early so it can be re-used to save stack space
     telemetry_event_t sev = {0};
-    ptelemetry_event_t ev = &sev;
 
-    u64 id = enter_exec_4_11(SP_EXECVE, ctx, ppid, luid, ev);
+    u64 id = enter_exec_4_11(SP_EXECVE, ctx, ppid, luid, &sev);
```
RafaelOrtizRC, line 489: I've had issues with this and the verifier a lot in the past, which is why there's that janky allocation and pointer assignment. But since you tested it, I'm willing to say it was probably an issue with older verifiers we don't need to worry about any more.

src/process-events.c
```diff
@@ -502,17 +485,18 @@ int BPF_KPROBE_SYSCALL(kprobe__sys_execve_4_11,
...
-    __builtin_memset(ev, 0, sizeof(telemetry_event_t));
-    u64 off = 0;
-    char br = 0;
-    READ_VALUE_N(ev, TE_EXEC_FILENAME, filename, 5);
+    sev.id = id;
+    sev.telemetry_type = TE_EXEC_FILENAME;
+    sev.u.v.value[0] = '\0';
+    sev.u.v.truncated = FALSE;
+
+    push_string(ctx, &sev, filename);
```
RafaelOrtizRC, line 499: Same here as the previous comment. Since it passed, we should be good. If we're watching for verification errors we should be alerted to any edge cases we support where this might fail.

RafaelOrtizRC (approved):
Some suggestions, but overall looks good. Thanks again for putting in the work to verify compatibility!

src/process-events.c
rc-andres: Yeah I considered it, especially for the `argv` program where we now have a lot of head room but decided to only try that after I add the code to address the race condition since that was going to add instructions already.

rc-andres:
I just finished a test on Ubuntu 16 running `4.11.0-13-generic` and the consistency check passes as expected.

PR #41, opened 2022-04-18
Send done signals

rc-andres:
https://redcanary.atlassian.net/browse/CWP-595

This adds a new event `TE_ENTER_DONE` that tells the user space when the enter point for a syscall (usually the kprobe) is done sending events. This will let us synchronize with `TE_RETCODE` so we know when both the enter and exit events are done and we can now construct the event fully. I didn't know what to call this new event, I am also partial to `TE_KPROBE_DONE` but I didn't use it because in the exit syscall program there is no kprobe vs kretprobe, they are all done in the same kprobe and I didn't know if that name would then cause confusion. 

I also added quite a bit of tribal knowledge that I had in the README so it is not just stored in my or (former) R&D devs' brains 😁 

Its sibling PR is: https://github.com/redcanaryco/cwp-plugins/pull/117 which can can be pulled to help test the changes in the ebpf programs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,17 @@ check all the sections at the same time.
 
 eBPF programs can branch (but not jump back!) so make sure to check
 that none of the branches go over the 4096 instructions limit.
+
+## License
+
+Our eBPF programs have a section:
+
+```c
+char _license[] SEC("license") = "GPL";
+```
+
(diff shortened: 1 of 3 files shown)


pinealservo (approved):
The additional docs are nice!  I think the changes all make sense.

PR #43, opened 2022-05-05
Handle exec and fork in threads

rc-andres:
Includes 4 fixes to the process programs:

* We were reporting on clones caused by thread's spawning. The goal of the programs was to report on new PIDs being created, new TIDs aren't useful to us so we will now intentionally not send clone events with the `CLONE_THREAD` flag.

* Execs within threads were not being reported properly due to the tid and pid being different between the kprobe and the kretprobe. This fixes it by only using the pid for exec* syscalls. 

* We were pointlessly reporting on failed unshare calls.

* We were pointlessly reporting on failed clone-like calls. 

These are each split into their own commit so it. may be easier to review that way.

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -195,10 +195,12 @@ static __always_inline int start_syscall(struct pt_regs *ctx, ptelemetry_event_t
     return 0;
 }
 
-// If the retcode < 0, returns true and sends a discard event. Else,
-// returns false.
+// Sends a discard event and returns true if the retcode in the ctx is
+// an error code.
 static __always_inline bool check_discard(struct pt_regs *ctx, ptelemetry_event_t ev)
 {
+    // check for < 0 rather than just non-zero because forks return a
+    // positive value on successes (the pid of the new process)
     int retcode = (int)PT_REGS_RC(ctx);
     if (retcode < 0) {
         ev->telemetry_type = TE_DISCARD;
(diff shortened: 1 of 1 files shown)


src/process-events.c
```diff
@@ -90,6 +90,19 @@ int kprobe__do_mount(struct pt_regs *ctx)
...
+// the current thread group. This is essentially just the tgid (user
+// space pid) duplicated as the tid == pid in the main thread. This is
+// useful in exec* syscalls since they cause non-main threads to be
+// immediately terminated and the pid will be equal to the tid during
+// the kretprobe even if it wasn't during the kprobe.
+static __always_inline u64 main_thread_pid_tid()
+{
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    u64 pid = pid_tgid >> 32;
+    return pid << 32 | pid;
```
rc-andres, line 103: we could in theory make this just return pid and have that be the key but since all the other syscalls use the pid_tgid "format" as the keys (and they have to) it would feel weird to have a padded 32 bit number mixed in with the other 64 bit numbers imo.

rc-vbundage (approved):
Looks good! Only concern I had was about `exit_clone3` being taken out/replaced until I realized `exit_clonex` was doing the same thing anyway. I appreciate the comments you added in as well. 

You could potentially clarify further why we need to check for negative return codes rather than non-zero return codes in `check_discard` but not sure how nice of a condensed explanation could be made out of it.

rc-andres:
> 

I added a comment clarifying (:

PR #47, opened 2022-05-17
add early exits if the address family is not AF_INET or AF_INET6

rc-andres:
In `ret_inet_csk_accept` and `ret_tcp_connect` we were not properly returning early if the address family was not for ipv4 or ipv6. This PR adds a check and an early return for those cases.


rc-dbogle (approved):
Looks great! Thanks for doing that

PR #48, opened 2022-05-25
Use LRU filter for network events

erikwilson-rc:
Add filter for repeated network events by comparing: protocol, pid, remote port, and address.

### DIFF
--- a/src/network-events.c
+++ b/src/network-events.c
@@ -52,11 +52,17 @@ static __always_inline int save_sock_ptr(struct pt_regs *ctx, void *map)
     return 0;
 }
 
-typedef struct
-{
-    u16 protocol_type;           // Something like IPPROTO_TCP or IPPROTO_UDP
-    u32 pid;
+// Network connections will only emit once per "shape", the shape
+// being the `network_event_key_t`. This is done to prevent
+// overloading userspace with repetitive network connections by the
+// same process (e.g., a chatty UDP connection). Not including UDP vs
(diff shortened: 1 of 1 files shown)


src/network-events.c
```diff
@@ -52,6 +52,47 @@ static __always_inline int save_sock_ptr(struct pt_regs *ctx, void *map)
     return 0;
 }
 
+typedef struct
+{
+    u16 protocol_type;           // Something like IPPROTO_TCP or IPPROTO_UDP
```
rc-andres, line 57: Since we are changing the `remote_port` to be 0 for UDP, and "0" is not a port that a remote connection would have, is storing the `protocol_type` necessary at all? Might help us shave a bit of size out of the key.

src/network-events.c
```diff
@@ -52,6 +52,47 @@ static __always_inline int save_sock_ptr(struct pt_regs *ctx, void *map)
...
+    u32 pid;
+    u16 remote_port;
+    ip_addr_t protos;
```
rc-andres, line 60: nit: we should probably order this in ascending size so the struct has less gaps and takes less memory.

rc-andres (changes requested):
A couple of suggestions to do before merging.

rc-andres (approved):

rc-vbundage (approved):

pinealservo (approved):
Seems pretty straightforward thanks to BPF_MAP_TYPE_LRU_HASH.

PR #49, opened 2022-06-14
Collapse process messages

rc-andres:
# Summary

All but exec* syscall events will now only trigger a single message in the perfmap. Exec* syscalls will trigger two (and exactly two!) syscalls. I left some TODOs in there on purpose for places that we may consider doing more things either in this PR or after (e.g., sending error events from the programs to user space so we can monitor it better?). 

# Why

There have been a few bugs in the ebpf programs due to synchronization of the messages (see introduction of `TE_ENTER_DONE`). For some reason though, we keep seeing these errors in the wild even though we have had trouble replicating them locally. Additionally, whenever some messages in the middle get dropped it is hard to for the consumer of the messages to know if the event still has enough data to be valid or not. (e.g., it could drop part of the exe path, or the args, or just the return code). Also, because we sent a message per argument, or per path segment, we were often sending a LOT of messages for just a single event which was not efficient. This PR collapses them all into one (and two for exec*) messages per syscall event. This makes it much easier to handle in userspace since it no longer needs to wait for "all the right messages", it will already have all the necessary data in the one (or two) messages. 

# How

For non exec* events, this was pretty easy: We make a map of "incomplete data" that gets created in the kprobe, then the kretprobe reads this map (pid + tid as the key), and sends the full event in the kretprobe. This also lets us check for the return code before we send any messages through the perfmap thus saving space on failed syscalls. 

For exec* events, this is a little hairier for a couple of reasons:
  * The arguments that are passed as pointers (argv, filename) can only be read in the kprobe as the executable image changes by the time the kretprobe is called making thus pointers "invalid" by that time (the pointers belong in the virtual memory of the old image). This means that we cannot simply save the pointers, but rather we have to copy these strings into something. 
  * The data size is flexible as we want to know things like exe path, cwd, and arguments. This means that we need more memory than the stack of eBPF allows us to have (512 bytes). 
 
To address these two issues with exec* syscalls, a per-cpu buffer was added with 32KB of memory allocated per buffer. This buffer is used to save data that can't fit in the stack, namely the strings. This buffer is also aware of how much of its data is actually used so we don't have to send the entire 32KB to user space, but instead only the amount that was filled by the program. Since a kprobe and a kretprobe don't have to be executed in the same CPU, this buffer cannot be used to send information from the kprobe to the kretprobe. The kprobe sends a message with just string data and an event id (similar to the old architecture), and then the kretprobe sends another message with the same event id and the rest of the data (i.e., retcode). In user space, it is sufficient to wait for both messages to come in, and there is no need to synchronize which one needs to come first or last, it just always needs both. 

# Assumptions

* A path sent to execve will always be at most `PATH_MAX` (4096 w/ null byte). If this assumption is not met then the path will be truncated. While an absolute path can go over `PATH_MAX`, everything I saw online seemed to indicate that paths sent to syscalls will respect `PATH_MAX`. 

* A path segment will be at most `NAME_MAX` (256 w/ null byte). If this assumption is not met then segments of a current working directory or the executable for execveat will be truncated. I think this is a good enough assumption and simplifies code both in the programs and in readers of the messages so it doesn't have to handle truncations. If this proves problematic we can bump it. 

* Each argument will be at most 256 bytes. Each argument will be truncated after that. I think this is preferred to having a single argument eat up all of the available buffer. At least this way it gives individual arguments enough space so we know what it was trying to do. If this proves problematic we can bump it. 

* Same as the old programs, we are restricted on number of tail calls and instructions. Arguments are limited to 1200 (75 per tail call and 16 tail calls). Path segments are limited to 204 (12 per tail call, and 17 tail calls). The code purposefully does not let the path segments from from tail calling further to give room for the arguments. The number of instructions per tail call is limited by the 4096 instruction limit in older kernels. We can potentially bump this further for newer kernels if we made separate programs, that's left as a todo for another time if necessary. 

* Same as the old programs, we are assuming that there is no more than 1024 syscalls being executed at one time (the size of the `incomplete_events` map). This number can be very easily bumped though. Hashes aren't pre-allocated (unlike arrays) so I don't think there is much cost to making this number higher, so I may bump it to something like 8K to be "extra safe" but I am not yet sure.

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -79,6 +79,22 @@ struct bpf_map_def SEC("maps/tail_call_table") tail_call_table = {
     .namespace = "",
 };
 
+typedef struct {
+    process_message_type_t type;
+    union {
+        int unshare_flags;
+    };
+} incomplete_event_t;
+
+struct bpf_map_def SEC("maps/incomplete_events") incomplete_events = {
+    .type = BPF_MAP_TYPE_HASH,
+    .key_size = sizeof(u64),
+    .value_size = sizeof(incomplete_event_t),
+    .max_entries = 1024,
+    .pinning = 0,
+    .namespace = "",
+};
+
 SEC("kprobe/do_mount")
 int kprobe__do_mount(struct pt_regs *ctx)
 {
@@ -180,7 +196,8 @@ static __always_inline int push_syscall(struct pt_regs *ctx, pprocess_message_t 
 }
 
 static __always_inline int send_syscall(struct pt_regs *ctx, pprocess_message_t ev,
-                                        process_message_type_t pm, void *ts, u64 pid_tgid)
+                                        process_message_type_t pm, void *ts, u64 pid_tgid,
+                                        process_data_union_t data)
 {
     if (!offset_loaded()) return -1;
     if (!is_user_process(ts)) return -1;
@@ -206,6 +223,7 @@ static __always_inline int send_syscall(struct pt_regs *ctx, pprocess_message_t 
         .euid = uid_gid >> 32,
         .egid = uid_gid & 0xFFFFFFFF,
         .mono_ns = bpf_ktime_get_ns(),
+        .data = data,
     };
 
     return push_message(ctx, ev);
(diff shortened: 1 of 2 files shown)


src/process-events.c
```diff
@@ -116,34 +167,49 @@ static __always_inline void* is_user_process(void *ts)
...
+    // smarter implementation as it complains about potential out of
+    // bounds or negative values. While perhaps theoretically possible
+    // to improve this with just the right incantations (and maybe
+    // turning off some compiler optimizaitons that remove some
+    // checks) at this time this is considered good enough (TM).
+
+    // already too full
+    if (*offset > MAX_PERCPU_BUFFER - max_string) return 0;
+
+    int sz = bpf_probe_read_str(&buffer->buf[*offset], max_string, string);
```
RafaelOrtizRC, line 201: This is the part I recall the verifier having the most trouble with. If you can get it to work, this is a much better way to send information for all the reasons you already listed.

rc-andres: Yeah that part gave me a LOT of issues when I was trying to be smart about it (hence the giant comment right above it). Now that the "max_string" is hardcoded then the verifier is happy as long as the bounds check is very close to it (the if statement right above it). I spent a good chunk of time trying to make it work with fancier code, even adding compiler barrier (asm volatile) and it just wouldn't like it so I gave up and made it dumb enough for the verifier to be happy with it

RafaelOrtizRC: Yep, that's the verifier.

rc-andres: I have tested that it works on:

* Ubuntu 16.04 4.15.0-112-generic
* Ubuntu 18.04 5.4.0-113-generic
* Centos 7 3.10.0-1160.53.1.el7.x86_64
* Centos 8 4.18.0-365.el8.x86_64

rc-vbundage (approved):
Looks good with me! Thanks for putting in those doc comments and detailing out the reasoning behind the changes in your description. 

I'm approving but I'd say wait for some others to take a look as well. 👍🏾

erikwilson-rc (approved):
🎉

PR #58, opened 2022-09-22
introduce make dev so the programs can be easily made in dev machines

rc-andres:
I think there's a question on whether it is valuable for us to keep building with such old clang + llvm. The tricky part and the part we cannot change is using the 4.11 kernel headers so we aren't adding breaking changes but building with old clang/llvm may actually be keeping us from fixes/optimizations in the compiler. Anyway this PR doesn't address that but I want us to be thinking about that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ SRC = src
 SRCS = $(wildcard $(SRC)/*.c)
 OBJS = $(patsubst $(SRC)/%.c,$(SRC)/%,$(SRCS))
 OBJS_WRAPPED = $(OBJS)
-CLANG_VER = 8
+CLANG_VER ?= 8
 CC = clang-$(CLANG_VER)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -27,6 +27,11 @@ CFLAGS += \
 	-Xclang -disable-llvm-passes \
 	-O2
 
+ifeq ($(shell test $(CLANG_VER) -gt 10; echo $$?),0)
+  CFLAGS += \
+	-Wno-frame-address
+endif
+
```
rc-andres, line 34: `Wframe-address` was introduced in clang 11 so for the sake of not breaking the default build of clang 8 let's keep this as is.

erikwilson-rc (approved):

rc-vbundage (approved):
🥳

PR #60, opened 2022-09-27
Use exe and args from task struct

rc-andres:
Alright this is one is a biggie:

We cannot rely on user input in the exec* syscalls for exename and argv. In particular, we cannot rely on this because the kernel overwrites them for cases such as interpreted scripts (e.g., bash scripts). So we are now going to rely on the data inside the `task_struct` in the `kretprobe` of exec* syscalls to gather all that data. This actually simplifies a lot of logic:

* `execveat` and `execve` are now exactly the same.
* We no longer need to pass incomplete data from kprobe to neither user space not to kretprobe for exec syscalls. All the information we want is in the `kretprobe` so there is nothing for the `kprobe` to do. 
* The kernel when it copies the `argv` it puts it all into a contiguous buffer and it knows the `arg_start` and `arg_end`. As far as I can tell the kernel itself relies on this: https://elixir.bootlin.com/linux/v4.11/source/mm/util.c#L629. Because of this we can now just copy the entire thing in one go rather than loop. This greatly improves our performance. This also helps us fix a bug we hadn't noticed before: because we relied on double null separators arguments of empty strings would make us stop the parsing early and not parse the entire argv. 

Additionally I added an improvement that let's us traverse our path trees without the skipping logic by keeping track of the dentry pointer itself.

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -44,11 +44,6 @@ typedef struct {
     clone_info_t clone_info;
 } incomplete_clone_t;
 
-typedef struct {
-    u64 pid_tgid;
-    u64 exec_id;
-} incomplete_exec_t;
-
 // used for events with flexible sizes (i.e., exec*) so it can send
 // extra data. Used in conjuction with a map such that it does not use
 // the stack size limit.
@@ -82,17 +77,6 @@ struct bpf_map_def SEC("maps/process_events") process_events = {
     .key_size = sizeof(u32),
     .value_size = sizeof(u32),
     .max_entries = 0, // let oxidebpf set it to num_cpus
-    .pinning = 0,
-    .namespace = "",
-};
-
-// A map of execs that have started (a kprobe) but are yet to finish
-// (the kretprobe).
-struct bpf_map_def SEC("maps/incomplete_exec") incomplete_execs = {
-    .type = BPF_MAP_TYPE_LRU_HASH,
-    .key_size = sizeof(u64),
-    .value_size = sizeof(incomplete_exec_t),
-    .max_entries = 8 * 1024,
     .pinning = 0,
     .namespace = "",
 };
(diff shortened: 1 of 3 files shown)


src/process-events.c
```diff
@@ -400,526 +349,289 @@ static __always_inline int write_path(struct pt_regs *ctx, void *ptr, buf_t *buf
...
+        // deliberately making a new process_message_t instead of
+        // reusing `pm`. The goal is to still push the truncated argv
+        // so we do not want to override any of the data in `pm`
+        process_message_t warning = {0};
+        warning.type = PM_WARNING;
+        warning.u.warning_info.pid_tgid = pid_tgid;
+        warning.u.warning_info.message_type = pm_type;
+        warning.u.warning_info.code = PMW_ARGV_TOO_BIG;
+        warning.u.warning_info.info.total_len = argv_len;
+        push_message(ctx, &warning);
```
pinealservo, line 460: Is there maybe room somewhere in `pm` to store a "truncated" flag instead of pushing a whole new message?

rc-andres: If we did that then I would want it to be a truncated and an original length field so we know that on the cases this may not be enough how much more it may need to be.

src/types.h
```diff
@@ -252,9 +254,12 @@ typedef struct
         clone_info_t clone_info;
         struct
         {
-            u64 event_id;
             u32 buffer_length;
             file_info_t file_info;
+            // send the length of argv we are processing. This is
+            // necessary because unlike paths, argvs can have empty
+            // strings so we cannot rely on double null separators
+            u32 argv_len;
```
pinealservo, line 262: This seems way more room than we need for the length; maybe a u16 length and u16 flags to store the truncated flag?

rc-andres: `u16` for the truncated length makes sense. I think technically the total length can be bigger than that, although I think a `u32` will do it. The number of arguments is limited to 2^31-1, and each argument is limited to `PAGE_SIZE*32`: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/binfmts.h#L16.

When actually doing the calculations then it also takes into account that the whole thing can't be more than 1/4 of stack size (defaulted to 8MB?) nor 3/4 or `_STK_LIM` (set to 8 MB). So if I am doing my math right, very very worse case combined argv + envp cannot be more than 6 MB, so less than 32 bits but more than 16 bits.

I'll change it with these values.

pinealservo: Well, the kernel may accept ridiculously sized ARGV arrays, but we only have 32KB of room in the buffer we're copying to.  I don't know that there's much to be gained by knowing how much got truncated beyond 32KB (there may be a lot more, but is it of any value to us?), but it probably doesn't hurt to store it instead of just a flag.

src/process-events.c
rc-andres: Done! (:

src/types.h
rc-andres: Changed it to use `u16` for the truncated length and then passing a flag saying it was truncated.

pinealservo (approved):
I think this looks good now; we might consider shrinking the remaining syscall context maps some more as well, but a bit of extra space shouldn't hurt too much at this point.

PR #62, opened 2022-10-21
handle recursive mounts and mountroots not at "/"

rc-andres:
I was seeing a weird path when running processes inside a mounted folder inside docker due to our lack of handling "mount roots" properly. 

We were also not handling nested mounts because we were only going one layer deep in the mount instead of looking at that mount's parent. 

The logic now is essentially a simplified version of: https://github.com/torvalds/linux/blob/247f34f7b80357943234f93f247a1ae6b6c3a740/fs/d_path.c#L226 (the main logic being in the helper function: https://github.com/torvalds/linux/blob/master/fs/d_path.c#L103)

Simplified in that it doesn't worry about only going up to an arbitrary "root" (it just goes until the absolute root and that it doesn't handle cases of an unattached namespace). 

The logic is summarized here:

#### Once when starting to process the path 
1. Save the `dentry` (`path->dentry`) and the `vfsmount` (`path->mnt`) of the path in a map that can be used for tail-calling. 

#### Once per tail-call

2. Calculate the `mount` that wraps the `vfsmount`.
3. Calculate the parent `mount` (`mount->mnt_parent`)
4. Calculate the  mount root dentry (`vfsmount->mnt_root`)

#### Once per path segment

5. If the current `dentry` is already at `vfsmount->mnt_root`:
    1. If our current `mount` is the same as our `mount->parent`:
        1. Done processing the path go to global root handling (Step 11).
    2. Change `dentry` to be `mount->mountpoint` as we will now start to process the mountpoint as our path
    3. Set `mount` to be `mount->parent` (also change the variable mnt_parent accordingly)
    4. Change `vfsmount` to be the new `mount->mnt` (this updates the across-tail call cache). Change the mount root dentry variable accordingly.
    5. Continue back from 5.
6. If `dentry == dentry->d_parent`:
    1. We know we aren't at the mount root so our path isn't related to the mount (e.g., it's a memfd). Go to global root  handling (Step 11)
7. Save the `dentry`'s path segment (this updates the across-tail call cache)
8. Set `dentry->d_parent` to be the current `dentry`
9. Continue back from 5

#### End of loop without finding the global root
10. Tail-call which will restart back from 2.

#### At global root
11. Write the root's path segment as we deliberately left the loop without processing the root segment. Expected values here at either `/` or the name of a memfd. 
12. Done

The important parts here is that we only stop when either the `mount`'s parent is itself (so we are the global root) or the dentry's parent is itself and it has no relation to the mountpoint (e.g., a memfd). This allows us to get the paths of multiple mountpoints. We also stop each mountpoint at its virtual fs mount root as that is how the path would be represented externally and how we would be able to find the path.

### DIFF
--- a/src/process-events.c
+++ b/src/process-events.c
@@ -48,14 +48,11 @@ typedef struct
 
 typedef struct
 {
-    // the dentry to the path currently being processed. At the
-    // beginning this is set to the top dentry of a path but it may be
-    // set to the path of the mount later
-    void *path;
-    // the virtual fs mount with which to find the mount root and the
-    // real mount
+    // the dentry to the next path segment
+    void *next_dentry;
+    // the virtual fs mount where the path is mounted
     void *vfsmount;
-} dentry_pair;
+} cached_path_t;
 
 struct bpf_map_def SEC("maps/mount_events") mount_events = {
     .type = BPF_MAP_TYPE_PERF_EVENT_ARRAY,
@@ -67,10 +64,10 @@ struct bpf_map_def SEC("maps/mount_events") mount_events = {
 };
 
 // A per cpu cache of dentry so we can hold it across tail calls.
-struct bpf_map_def SEC("maps/percpu_dentries") percpu_dentries = {
+struct bpf_map_def SEC("maps/percpu_path") percpu_path = {
     .type = BPF_MAP_TYPE_PERCPU_ARRAY,
     .key_size = sizeof(u32),
-    .value_size = sizeof(dentry_pair),
+    .value_size = sizeof(cached_path_t),
     .max_entries = 1,
     .pinning = 0,
     .namespace = "",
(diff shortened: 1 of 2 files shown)


src/process-events.c
```diff
@@ -448,59 +439,73 @@ static __always_inline int write_path(struct pt_regs *ctx, dentry_pair *dentries
...
+    if (offset == NULL) return -1;
+    u32 mnt_root_offset = *(u32 *)offset; // offset of vfsmount->mnt_root
+
+    offset = get_offset(CRC_MOUNT_MOUNTPOINT);
+    if (offset == NULL) return -1;
+    u32 mountpoint_offset = *(u32 *)offset; // offset of mount->mountpoint
+
+    offset = get_offset(CRC_MOUNT_MNT);
+    if (offset == NULL) return -1;
+    u32 mnt_offset = *(u32 *)offset; // offset of mount->mnt
```
rc-andres, line 458: all of these offsets are retrieved in this way rather than our helper methods because we want to "cache" the offset values instead of getting them from the map on every iteration of the loop

pinealservo: It's kind of gross, but until we can do load-time relocations it's probably a reasonable tradeoff.

pinealservo (commented):
I have a question that wasn't clear to me from the code and my current level of understanding of how mount namespaces work in the kernel:

Is this going to follow mounts in containers all the way back to the outermost namespace's root, or does the pivot_root of a container cause its mounts to be rooted at the inner namespace's root? I'm guessing from my understanding of how mount namespaces work that it's the latter, but I'm not 100% sure.

pinealservo (approved):

rc-andres:
> I have a question that wasn't clear to me from the code and my current level of understanding of how mount namespaces work in the kernel:
> 
> Is this going to follow mounts in containers all the way back to the outermost namespace's root, or does the pivot_root of a container cause its mounts to be rooted at the inner namespace's root? I'm guessing from my understanding of how mount namespaces work that it's the latter, but I'm not 100% sure.

It is not going to go across namespaces, so the path will be from the perspective of the container and not the host.

PR #64, opened 2022-11-14
Split process event files

rc-andres:
This should be a no-op. I am just moving a lot of the existing logic into separate files as a precursor to adding even more logic for script events being included into process events.


pinealservo (approved):
Looks like it will help make things a bit easier to follow.

PR #65, opened 2022-11-17
Adding scripts as part of process messages

rc-andres:
The goal of this PR is to add script information into the messages produced by exec* syscalls. This is unlike our current script events that puts them into their own perfmap buffer as their own separate events. We want to have them be part of the "process events" so they can be easily synchronized. 

This PR adds:

Three new maps:

* `interpreters`: A file identity -> *absolute* interpreter path. This map is used to send the interpreter paths across probes. Because this is not keyed by a pid it allows for more efficient storage (e.g., only one entry for /bin/bash instead of one per bash script). This is only meant to store paths from scripts where the interpreter line is given as an absolute line (e.g., `#!/bin/bash`). This map does not handle cache invalidations caused by file deletions/renames and the reuse of inodes instead it relies on being rewritten everytime we encounter an interpreter and only used within the span of the same syscall thus making any race conditions extremely unlikely. 

* `rel_interpreters`: A file identity *+ pid* -> *relative* interpreter path. This map is meant to store paths for interpreters that were given in a relative path (e.g., `#!./my/customer/interpreter`). Because paths are relative the key also includes the *pid* so userspace can concatenate the cwd of a process to the relative path of the interpreter. The expectation is that this map will be rarely used/needed hence its small number of entries. 

* `scripts`: A map of PID -> script. This map is used to send script data across probes for a currently running exec. The key is the pid instead of the pid+tid because execs can start executing in any arbitrary thread but will always finish in the main thread (stopping all the other threads) so we cannot rely on the tid being the same from the script load probe to the exec kretprobe. 

Three new kprobes:

* `execve` and `execveat`: these new probes exist solely to clean up any dirty data from previous scripts caused by `kretprobe` not firing due to the max active limit (2 * # of cpus).
* `load_script`: this probe collects the necessary data and starts to build the different scripts + interpreters that were used for the exec* syscall saving it in a map that is cached by the user space PID (kernel TGID).

The logic is:
   1. If there is no saved data for that pid in the `scripts` map then we know it is the first "script load". We initialize (and save) a `script` instance with the file information (relative path + device + inode) and no interpreters. This is the easiest case because the relative path for the file is saved under `bprm->filename` which is a pointer to a kernel allocated string that is not de-allocated for the duration of the exec so we don't need to copy it we can just save the pointer to the string. At this point we do not yet know if we are dealing with a script or not, all (successful) execs call `load_script` at least once. 
   
   2. If there is already saved data then the we must be in the second or later call to "script load" so we are now dealing with interpreters. Unlike `bprm->filename`, `bprm->interp` (the path to the interpreter) is de-allocated in between calls to `load_script` so we need to save it somewhere. If `interp` is an absolute path (starts with `/`) then we save it in the `interpreters` map. If `interp` is a relative path then we save it in the `rel_interpreters` map with the current PID. For either map we always override what's in there if the key already exists so we don't have to worry about inodes being reused because something was removed. 
   
   3. In the `kretprobe` of exec* we check if there are any interpreters in the entry for `scripts` for the current PID. If there are no scripts information, or if the script information doesn't have any interpreters then we exit early. This is expected behavior (e.g., script program was not loaded or the execution was for not a script). If there are any interpreters then we add the filename of the original script and add all but the last interpreter path. We skip the last interpreter because that is the executable not a script so it doesn't belong in this message (we'll get information for it in the exec* message). In most cases we expect there to be only one interpreter so we won't even be looking up any interpreter paths at all. In the rare case we do need to lookup interpreter paths then we first check `rel_interpreters` in case we need to handle relative paths, and if not there we use the more expected case of `interpreters`. If there is no information in either that means our path got evicted and we enter an empty path and continue. We deliberately still emit the script information because we still have valuable data with the original script filename. 
   
   4. We create an `event_id` that is our pid (same as our tid because we are in an exec) combined with a random number. This event_id will be used for both the script message and the exec message so the messages can be combine into a single event in userspace. Because this is all happening in the same `kretprobe` we are avoiding any re-ordering issues. If a script wasn't found or if something failed we return 0 as an `event_id` to signal that the exec message won't have a script message waiting for it.


rc-andres:
Things to do:

* Change all the `bpf_printk` w/ proper error handling (e.g., emit warnings) or a comment explaining why we don't need to handle that case
* Handle interpreters using relative paths better. Right now there is a subtle bug I only just realized as I wrote the description where if two scripts are run "at the same time" at the same time pointing to the same file but using a relative rather than an absolute path and the relative paths are different (because the cwd are different) then the paths are not guaranteed to be correct for the emitted data as they will be relative paths from the incorrect cwd. I think things like this would fairly rare but just to be safe I may add a new map with less elements to handle relative paths using PID as part of the key. I still want this "larger" map that relies solely on file identity for absolute paths because it allows us to-use entries easily for the common case (running scripts using a system installed interpreter such as bash or ruby or python).

src/types.h
```diff
@@ -351,6 +354,10 @@ typedef union
 {
     syscall_info_t syscall_info;
     warning_info_t warning_info;
+    struct {
+        file_info_t scripts[4];
```
rc-dbogle, line 358: Does this 4 correspond to MAX_INTERPRETERS?

rc-andres: Yes and no. This is (1 script filename) + (MAX_INTERPRETERS - 1). The -1 is because we don't care about the last interpreter because that is the same as the executable.

src/process/script.h
```diff
@@ -0,0 +1,257 @@
...
+
+  file_info_t interpreters[MAX_INTERPRETERS];
+} script_t;
+
+// A map containing script information for currently running exec*
+struct bpf_map_def SEC("maps/scripts") scripts = {
+  .type = BPF_MAP_TYPE_LRU_HASH,
+  .key_size = sizeof(u32),
+  .value_size = sizeof(script_t),
+  .max_entries = 1024, // TODO: should we expand oxidebpf to allow scaling based on # cpus?
```
pinealservo, line 30: IMO yes, being able to be nicer to smaller low-core-count systems while still being able to handle high-core-count ones seems worth a bit of effort to me. But probably for after the basic implementation is working.

src/process/script.h
```diff
@@ -0,0 +1,257 @@
...
+      intp_path = bpf_map_lookup_elem(&interpreters, intp_key);
+    }
+
+    if (intp_path == NULL) {
+      // we couldn't find the saved path given the file identity. This
+      // most likely means we got a *lot* of interpreter scripts in
+      // quick succession and made this one be evicted from our LRU
+      // cache. This is not an error so just write a null character to
+      // tell userspace that we don't have a path for this interpreter
+      // and then move on. Do not exit early because we still got
```
pinealservo, line 229: What have we still got?  The comment has left me hanging...

src/process/script.h
```diff
@@ -0,0 +1,257 @@
...
+    pm->u.script_info.scripts[i] = *intp_key;
+
+    relative_file_info_t rel_interpreter = {0};
+    rel_interpreter.pid = pid;
+    rel_interpreter.interpreter = *intp_key;
+
+    interpreter_path_t *intp_path = bpf_map_lookup_elem(&rel_interpreters, &rel_interpreter);
+    if (intp_path == NULL) {
+      intp_path = bpf_map_lookup_elem(&interpreters, intp_key);
+    }
```
pinealservo, line 221: It seems to me that we're far more likely to find it in the `interpreters` map rather than `rel_interpreters`; does it make sense to check `interpreters` first so we won't have to perform 2 map lookups as often?

src/process-events.c
```diff
@@ -173,5 +171,41 @@ int BPF_KPROBE_SYSCALL(kprobe__sys_exit_group, int status)
...
 
+SEC("kprobe/sys_execveat")
+int BPF_KPROBE_SYSCALL(kprobe__sys_execveat,
+                       int fd, const char __user *filename,
+                       const char __user *const __user *argv,
+                       const char __user *const __user *envp,
+                       int flags)
+{
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    u32 pid = pid_tgid >> 32;
```
pinealservo, line 182: I hate that `pid` means something different in the kernel than userspace.  I also hate that the names for these dual-32-bit helpers seem reversed from the layout of the value they return.

IIRC `pid_tgid >> 32` actualy gives us `tgid` which is what userspace considers the `pid`, but I wonder if we might call it some third thing instead so I don't end up second-guessing myself every time I read through these.  I can't think of anything equally short and clear though, so I'll leave it to your judgement.

src/process/exec.h
```diff
@@ -55,12 +31,14 @@ static __always_inline void exit_exec(struct pt_regs *ctx, process_message_type_
...
 
+    pm->u.syscall_info.data.exec_info.event_id = push_scripts(ctx, buffer);
+
     void *ts = (void *)bpf_get_current_task();
     u64 pid_tgid = bpf_get_current_pid_tgid();
 
     // do not emit if we couldn't fill the syscall info
     ret = fill_syscall(&pm->u.syscall_info, ts, pid_tgid >> 32);
-    if (ret > 0) return;
+    if (ret > 0) return; // TODO: send discard event if pm->u.syscall_info.data.exec_info.event_id != 0?
```
pinealservo, line 41: What's the idea behind these "send discard event?" comments?  Have we already sent something that we later need to ignore?

src/process/helpers.h
```diff
@@ -117,3 +118,26 @@ static __always_inline int fill_syscall(syscall_info_t *syscall_info, void *ts,
 
     return 0;
 }
+
+// it's argument should be a pointer to a file
```
pinealservo, line 122: Silly pedantic comment:  The apostrophe technically shouldn't be here; the rules for both marking contractions and marking possession call for it, but the disambiguation rule for this situation says that the contraction wins the apostrophe.

pinealservo (approved):
I did have some suggestions, but I don't think they're critical and you can apply them at your discretion.  I would like to hear your thoughts on the questions and comments I left though.

src/process/script.h
rc-andres: My only concern with that is that it leaves us potentially open to a very unlikely but possible error case:

1. A process starts a script with an interpreter in path `/my/absolute/path` with file identity X. Because the path is absolute it gets saved inside `/my/absolute/path`. This works as expected.
2. `/my/absolute/path` is renamed to `/my/absolute/interpreter` and the kernel happens to reuse the inode. We don't invalidate the path in the `interpreters` map because we haven't implemented that yet.
3. A process starts a script with cwd `/my/absolute` and the interpreter is `./interpreter`. Because it is relative the path gets saved inside `rel_interpreters` instead of `interpreters`. If we first look for it inside `interpreters` then we'll find the *old* path which is *incorrect*. 

So even though we are more likely to find the path inside `interpreters` rather than `rel_interpreters` if the file identity exists in both for a given PID then the one in `rel_interpreters` takes priority because it is more guaranteed to be newer. 

That being said I just spotted an issue with the current flow because `rel_interpreters` was not being invalidated so I just pushed a change to remove `rel_interpreters` when a new script is run using an absolute path for the same file identity + pid (e.g., double exec). Pretty unlikely to happen but just in case.

src/process/exec.h
rc-andres: At this point we have already sent the script information (if there was any to send) which the userspace would cache. I wonder if we should send discard events here or just let user-space age those out through some form of LRU cache or something. 

I don't think it's a big deal because we only send script events after we check the retcode of exec so the only things that could fail are if the exec was a "kernel process" which I don't think we get for execs for (much less execs with scripts) or errors in the ebpf program (e.g., missing offsets) which so far I have only seen happen during shutdown of a program. 

The TODO in here is mostly so I know "we'll probably want to address this later just to make it more fool-proof".

src/process-events.c
rc-andres: I agree it is pretty confusing. Right now what I've been doing is that as soon as the field is "translated" by doing the bitshift or mask I start naming them in terms of user-space. So in `u32 pid = pid_tgid >> 32`, `pid` is the user space process id (kernel space thread group id). In other spots where we do `u32 tid = pid_tgid & 0xFFFFFFFF`, `tid` is the user space threads id (kernel space process id).

src/process/script.h
rc-andres: Oops it's meant to be: "we still got useful information". I've updated the comment.

src/process/script.h
pinealservo: Based on my experimentation so far with filemod events, renaming via `renameat2` (which is what you get with `mv`) does re-use the inode, and you can also see inode re-use fairly easily with deletion followed by creation of a new item. So this all seems reasonable to me.

pinealservo (approved):
Again, seems to be in good shape. Just had a question I'm curious about; doesn't seem like it's a problem.

src/process/script.h
```diff
@@ -168,11 +148,41 @@ static __always_inline void enter_script(struct pt_regs *ctx, void *bprm) {
...
+    info.err = ret;
+    set_local_warning(PMW_UPDATE_MAP_ERROR, info);
+    goto EmitWarning;
+  }
+
   return;
 
+ EventMismatch:;
+  error_info_t info = {0};
+  info.stored_pid_tgid = (((u64) event.pid) << 32) & (event.pid);
```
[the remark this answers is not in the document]
rc-andres: Yeah we already use this type for mismatches when loading clone or unshare events which use the full 64-bit.

PR #66, opened 2022-12-01
[LE-516] fix bug where UID and GID were swapped

raidancampbell:
Tested locally and confirmed working.

as-is:
Screen Shot 2022-12-01 at 3 13 08 PM

to-be:
Screen Shot 2022-12-01 at 3 12 53 PM


pinealservo (approved):
I guess this is one of those bugs where getting the dev env and local build set up took 10x the time of the actual fix. :)

PR #69, opened 2023-01-14
fix load_script program to support execveat

rc-andres:
When a program is launched using an fd instead of solely a path the "filename" the binprm gets is a temporarily allocated string of the form `/dev/fd/*`. This string is de-allocated before the the kretprobe of exec* so reading that "filename" causes us to read garbagae data (although I don't think it can cause UB since the probe_read_str function will just fail if it tries to read uninitialized data). To avoid this, whenever we see a filename that starts with `/dev/fd/` we will copy the string into our interpreter path so that we can read it in the kretprobe safely. We will only do this for the /dev/fd/* case to still optimize for the common case

### DIFF
--- a/src/process/script.h
+++ b/src/process/script.h
@@ -124,7 +124,7 @@ static __always_inline void enter_script(struct pt_regs *ctx, void *bprm) {
   ret = bpf_probe_read_str(&path.path, BINPRM_BUF_SIZE, interp);
   if (ret < 0) {
     set_empty_local_warning(PMW_READ_PATH_STRING);
-    return;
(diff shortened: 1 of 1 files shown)


rc-andres:
For context: https://elixir.bootlin.com/linux/latest/source/fs/exec.c#L1513

pinealservo (approved):
Very cool; I wonder if a similar technique will be needed in picking up some of the filenames that we currently lose in audit during filemod syscalls (e.g. when you delete a directory and the name passed from userspace ends with '/').

PR #73, opened 2023-05-23
Initial filemod support monitoring mkdir calls

pinealservo:
This contains a lot of relatively minor refactoring to make some code written for process events also work for file events. Some more re-organization to move the shared code to a better location is coming.

The file event code is just a starting point, but an event type is in place and it successfully fills out create events when mkdir is used.

### DIFF
--- a/src/file-events.c
+++ b/src/file-events.c
@@ -133,9 +133,10 @@ static __always_inline void store_dentry(struct pt_regs *ctx, void *dentry)
     u64 pid_tgid = bpf_get_current_pid_tgid();
 
     load_event(incomplete_mkdirs, pid_tgid, incomplete_mkdir_t);
-    if (event.target_dentry == NULL)
+    if (event.target_dentry == NULL) {
         event.target_dentry = dentry;
-    bpf_map_update_elem(&incomplete_mkdirs, &pid_tgid, &event, BPF_ANY);
+        bpf_map_update_elem(&incomplete_mkdirs, &pid_tgid, &event, BPF_ANY);
+    }
     return;
 
     EventMismatch:;
--- a/src/file-events.c
+++ b/src/file-events.c
@@ -12,7 +12,7 @@
 #include "process/path.h"
 #include "process/warning.h"
 
-// The map where process event messages get emitted to
+// The map where file event messages get emitted to
 struct bpf_map_def SEC("maps/file_events") file_events = {
     .type = BPF_MAP_TYPE_PERF_EVENT_ARRAY,
     .key_size = sizeof(u32),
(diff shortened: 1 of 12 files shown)


src/file-events.c
```diff
@@ -0,0 +1,224 @@
...
+
+static __always_inline void store_dentry(struct pt_regs *ctx, void *dentry)
+{
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    error_info_t info = {0};
+
+    load_event(incomplete_mkdirs, pid_tgid, incomplete_mkdir_t);
+    event.target_dentry = dentry;
+    bpf_map_update_elem(&incomplete_mkdirs, &pid_tgid, &event, BPF_ANY);
+
```
rc-vbundage, line 139: Add a return!

src/file-events.c
```diff
@@ -0,0 +1,224 @@
...
+    EventMismatch:
+    info.stored_pid_tgid = event.pid_tgid;
+    set_local_warning(W_PID_TGID_MISMATCH, info);
```
rc-vbundage, line 142: We'll want to a push a warning here.

src/file-events.c
```diff
@@ -0,0 +1,224 @@
...
+    return;
+}
+
+static __always_inline void store_dentry(struct pt_regs *ctx, void *dentry)
+{
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    error_info_t info = {0};
+
+    load_event(incomplete_mkdirs, pid_tgid, incomplete_mkdir_t);
+    event.target_dentry = dentry;
```
rc-vbundage, line 137: we should check that this is NULL to prevent overwriting when using overlayfs

src/file-events.c
```diff
@@ -0,0 +1,232 @@
...
+#include "bpf_helpers.h"
+#include "types.h"
+#include "offsets.h"
+#include "common.h"
+#include "process/helpers.h"
+#include "process/buffer.h"
+#include "process/path.h"
+#include "process/warning.h"
+
+// The map where process event messages get emitted to
```
rc-vbundage, line 15:
```suggestion
// The map where file event messages get emitted to
```

src/file-events.c
```diff
@@ -0,0 +1,232 @@
...
+    ResolveName:
+    ret = write_path(ctx, cached_path, &cursor, tail_call);
+    cached_path->next_dentry = NULL;
+    if (ret < 0) goto EmitWarning;
+
+    write_null_char(cursor.buffer, cursor.offset);
+    push_flexible_file_message(ctx, fm, *cursor.offset);
+    return;
+
+    EventMismatch:
```
rc-vbundage, line 195: question: Does this need to be `EventMismatch:;` (end with semicolon)?

pinealservo: It's not normal to put a semicolon after a label; I think it was done in some places to allow a variable declaration immediately after the label, which isn't normally allowed.

rc-andres (approved):

PR #74, opened 2023-06-09
[LE-727] Hook up path filter state machine

pinealservo:
This adds the path lookup mechanism using the filter state machine to the path resolution code.

Additionally, filemod path lookups were previously broken due to not including the vfsmount field in the cached_path, which caused lookup to stop at the first mount boundary. This is now fixed, which required hooking a different one of the security functions so we could get a path instead of just a dentry.

### DIFF
--- a/src/file-events.c
+++ b/src/file-events.c
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0+
 
+// Configure path.h to include filter code 
+#define USE_PATH_FILTER 1 
+
 #include <linux/kconfig.h>
 #include <linux/version.h>
 #include <linux/fs.h>
@@ -55,10 +58,9 @@ static __always_inline int push_file_warning(struct pt_regs *ctx, file_message_t
     return push_file_message(ctx, fm);
 }
 
-// its argument should be a pointer to a dentry
-static __always_inline int extract_file_info_owner(void *ptr, file_info_t *file_info, file_ownership_t *file_owner)
+static __always_inline int extract_file_info_owner(void *dentry, file_info_t *file_info, file_ownership_t *file_owner)
 {
-    void *d_inode = read_field_ptr(ptr, CRC_DENTRY_D_INODE);
+    void *d_inode = read_field_ptr(dentry, CRC_DENTRY_D_INODE);
     if (d_inode == NULL) {
         // TODO: Once we are confident we can ignore dentries without inodes, return a "filter me" value
         // Right now we are only aware of this happening with mkdir in cgroupfs
@@ -100,8 +102,8 @@ static __always_inline int extract_file_info_owner(void *ptr, file_info_t *file_
 typedef struct {
   u64 pid_tgid;
   u64 start_ktime_ns;
-  void *target_vfsmount;     // vfsmount of the containing directory
-  void *target_dentry;  // dentry of the new directory
+  void *target_vfsmount;  // vfsmount of the containing directory
+  void *target_dentry;    // dentry of the new directory
 } incomplete_mkdir_t;
 
 // A map of mkdirs that have started (a kprobe) but are yet to finish
@@ -133,27 +135,24 @@ static __always_inline void enter_mkdir(struct pt_regs *ctx)
         fm.u.warning.info.err = ret;
 
         push_file_message(ctx, &fm);
-
-        return;
     }
-
-    return;
 }
 
 static __always_inline void store_dentry(struct pt_regs *ctx, void *path, void *dentry)
 {
     u64 pid_tgid = bpf_get_current_pid_tgid();
+    file_message_t fm = {0};
 
     load_event(incomplete_mkdirs, pid_tgid, incomplete_mkdir_t);
     if (event.target_dentry == NULL) {
         event.target_dentry = dentry;
         event.target_vfsmount = read_field_ptr(path, CRC_PATH_MNT);
+        if (event.target_vfsmount == NULL) goto EmitWarning;
         bpf_map_update_elem(&incomplete_mkdirs, &pid_tgid, &event, BPF_ANY);
     }
     return;
 
-    EventMismatch:;
-    file_message_t fm = {0};
+    EventMismatch:
     fm.type = FM_WARNING;
     fm.u.warning.pid_tgid = pid_tgid;
     fm.u.warning.message_type.file = FM_CREATE;
@@ -162,6 +161,9 @@ static __always_inline void store_dentry(struct pt_regs *ctx, void *path, void *
 
     push_file_message(ctx, &fm);
     return;
+
+    EmitWarning:
+    push_file_warning(ctx, &fm, FM_CREATE);
 
     NoEvent:
     return;
@@ -205,11 +207,9 @@ static __always_inline void exit_mkdir(struct pt_regs *ctx, tail_call_slot_t tai
     if (ret < 0) goto EmitWarning;
     if (cached_path->filter_state <= 0) goto NoEvent; // Didn't match watched path filter
 
-    bpf_probe_read_str(fm->u.action.filter_tag, MAX_TAG, cached_path->filter_tag);
+    fm->u.action.tag = cached_path->filter_tag;
     push_flexible_file_message(ctx, fm, *cursor.offset);
-    // My attempts to call bpf_map_delete_elem() so far have caused the verifier to fail it
-    // so just rely on the LRU functionality to keep the map from getting too big
-    write_null_char(cursor.buffer, cursor.offset);
+
     // lookup tail calls completed; ensure we re-init cached_path next call
     cached_path->next_dentry = NULL;
     return;
@@ -222,6 +222,7 @@ static __always_inline void exit_mkdir(struct pt_regs *ctx, tail_call_slot_t tai
     push_file_warning(ctx, fm, FM_CREATE);
 
     NoEvent:
+    // lookup tail calls completed; ensure we re-init cached_path next call
     cached_path->next_dentry = NULL;
     return;
 }
--- a/src/file-events.c
+++ b/src/file-events.c
@@ -100,7 +100,8 @@ static __always_inline int extract_file_info_owner(void *ptr, file_info_t *file_
 typedef struct {
   u64 pid_tgid;
   u64 start_ktime_ns;
-  void *target_dentry;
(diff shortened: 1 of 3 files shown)


src/file-events.c
```diff
@@ -139,13 +140,14 @@ static __always_inline void enter_mkdir(struct pt_regs *ctx)
...
 
-static __always_inline void store_dentry(struct pt_regs *ctx, void *dentry)
+static __always_inline void store_dentry(struct pt_regs *ctx, void *path, void *dentry)
 {
     u64 pid_tgid = bpf_get_current_pid_tgid();
 
     load_event(incomplete_mkdirs, pid_tgid, incomplete_mkdir_t);
     if (event.target_dentry == NULL) {
         event.target_dentry = dentry;
+        event.target_vfsmount = read_field_ptr(path, CRC_PATH_MNT);
```
rc-andres, line 150: `read_field_ptr` purposefully sets the local warning if the pointer is `NULL`. I think we always expect `vfsmount` not to be NULL so we should emit a warning if we get into a bad state where it is NULL somehow and then exit early out of here.

src/file-events.c
```diff
@@ -139,13 +140,14 @@ static __always_inline void enter_mkdir(struct pt_regs *ctx)
...
 }
 
-static __always_inline void store_dentry(struct pt_regs *ctx, void *dentry)
+static __always_inline void store_dentry(struct pt_regs *ctx, void *path, void *dentry)
 {
     u64 pid_tgid = bpf_get_current_pid_tgid();
 
     load_event(incomplete_mkdirs, pid_tgid, incomplete_mkdir_t);
     if (event.target_dentry == NULL) {
         event.target_dentry = dentry;
```
rc-andres, line 149: should we early guard against this being NULL? I know we check for it being NULL later but I am wondering if we should be checking it before putting it into the `incomplete_mkdirs` at all.

src/file-events.c
```diff
@@ -196,17 +198,20 @@ static __always_inline void exit_mkdir(struct pt_regs *ctx, tail_call_slot_t tai
...
+    init_filtered_cached_path(cached_path, event.target_dentry, event.target_vfsmount);
 
     ResolveName:
     ret = write_path(ctx, cached_path, &cursor, tail_call);
-    cached_path->next_dentry = NULL;
     if (ret < 0) goto EmitWarning;
+    if (cached_path->filter_state <= 0) goto NoEvent; // Didn't match watched path filter
 
-    write_null_char(cursor.buffer, cursor.offset);
+    bpf_probe_read_str(fm->u.action.filter_tag, MAX_TAG, cached_path->filter_tag);
```
rc-andres, line 208: should we emit a warning if this fails?

src/file-events.c
```diff
@@ -196,17 +198,20 @@ static __always_inline void exit_mkdir(struct pt_regs *ctx, tail_call_slot_t tai
...
     push_flexible_file_message(ctx, fm, *cursor.offset);
     // My attempts to call bpf_map_delete_elem() so far have caused the verifier to fail it
     // so just rely on the LRU functionality to keep the map from getting too big
+    write_null_char(cursor.buffer, cursor.offset);
```
rc-andres, line 212: why do we write a null char into the buffer here? This buffer isn't going to be used anymore in this probe afaict and since it's per-cpu any subsequent probe should restart the cursor from the start.

src/process/path.h
```diff
@@ -154,6 +199,25 @@ static __always_inline int write_path(struct pt_regs *ctx, cached_path_t *cached
...
+            // One retry with "*" path element if not found
+            if (val == NULL) {
+                // The retry_key always uses the same path segment string so we don't need to reset
+                retry_key.current_state = cached_path->filter_state;
+                val = (filter_value_t *)bpf_map_lookup_elem(&filter_transitions, &retry_key);
+            } 
+            // Terminate the filter lookup if there was no value at either key
+            cached_path->filter_state = val ? val->next_state : -1;
+        }
+
```
rc-andres, line 220: This code is re-used in the execve* kretprobes which were already pretty close to its max instructions and now they are going over  (even though it will never hit that if but the verifier doesn't know that). The easiest "fix" would be to reduce the number of segments we can handle per tail (`MAX_PATH_SEGMENTS_NOTAIL`) which is currently at 25 but it'd be cool if we could find different way

src/process/path.h
```diff
@@ -154,6 +199,25 @@ static __always_inline int write_path(struct pt_regs *ctx, cached_path_t *cached
         if (bpf_probe_read(&offset, sizeof(offset), dentry + name) < 0)
             goto NameError;
 
+        // Path filtering, only if the filter is active
+        if (cached_path->filter_state >= 0) {
+            val = NULL;
+            // We need to reset at least the whole string storage array each time we use the key
+            bpf_probe_read(&key, sizeof(filter_key_t), &blank_key);
```
rc-andres, line 206: is this easier/less instructions than a `key = (filter_key_t){0}`?

src/process/path.h
```diff
@@ -154,6 +199,25 @@ static __always_inline int write_path(struct pt_regs *ctx, cached_path_t *cached
...
+                val = (filter_value_t *)bpf_map_lookup_elem(&filter_transitions, &key);
+            }
+            // One retry with "*" path element if not found
+            if (val == NULL) {
+                // The retry_key always uses the same path segment string so we don't need to reset
+                retry_key.current_state = cached_path->filter_state;
+                val = (filter_value_t *)bpf_map_lookup_elem(&filter_transitions, &retry_key);
+            } 
+            // Terminate the filter lookup if there was no value at either key
+            cached_path->filter_state = val ? val->next_state : -1;
```
rc-andres, line 218: for the cases where we *want* to do the filtering, after this returns `-1` any future iteration/tail call isn't needed (afaict) so should we be exiting early here? Maybe return something out of `write_path` that indicates "stopped because got filtered out"?

src/types.h
```diff
@@ -377,13 +377,26 @@ typedef struct
     char strings[];
 } process_message_t, *pprocess_message_t;
 
+#define MAX_PATH_SEG 64
```
rc-andres, line 380: just so I understand this, this is just saying that the paths we allow to filter on must be less than 64 characters right? There could be a segment in the path that is bigger than that but that segment just won't be specified in the filter.

src/file-events.c
```diff
@@ -196,17 +198,20 @@ static __always_inline void exit_mkdir(struct pt_regs *ctx, tail_call_slot_t tai
     fm->u.action.buffer_len = sizeof(file_message_t);
     fm->u.action.u.create.source_link = LINK_NONE;
 
-    cached_path->next_dentry = event.target_dentry;
+    init_filtered_cached_path(cached_path, event.target_dentry, event.target_vfsmount);
 
     ResolveName:
     ret = write_path(ctx, cached_path, &cursor, tail_call);
-    cached_path->next_dentry = NULL;
```
rc-andres, line 203: I think we should keep this here. I'd consider it safer to reset `next_dentry` immediately. Otherwise in the code as it is written right now it the case of an error in `write_path` it doesn't look like we are resetting `next_dentry` so the next time this CPU loads `percpu_path` it will still have a `next_dentry` in there.

src/process/path.h
```diff
@@ -11,8 +11,22 @@ typedef struct
     void *next_dentry;
     // the virtual fs mount where the path is mounted
     void *vfsmount;
+    // filter state machine state; init to 0 for filtering, -1 for none
+    int filter_state;
+    // filter match tag; filled by filter when a match is found
+    char filter_tag[MAX_TAG];
```
rc-andres, line 17: maybe I am misunderstanding something but these tags don't need to be strings right? When userspace sets these tags it could just be a number that is meaningful to userspace. If userspace wants it as a string later on for tags they could have a separate conversion from "integer tags" to the "named tags". I don't think it matters all that much but it may make the kernel side be just a tad simpler (just a u16 or something) 🤷🏽 .

src/process/path.h
pinealservo: Since helper functions aren't inlined, I suspect it's far fewer instructions to call a helper than write 0s into the whole struct.  Would have to compare generated code to be sure, though.

src/process/path.h
pinealservo: I did some experiments with early exiting but I wanted to verify the logic before trying to take extra shortcuts. I think in practice you don't end up getting stuck in the state machine like that very often before the root, so it might not save much.

src/types.h
pinealservo: Yes; we should still be able accept longer path segments via the wildcard matchers, but we can't distinguish path segments that don't differ until after their 63rd character in our globs.

src/file-events.c
pinealservo: There are only 2 `return` statements reachable after `write_path` returns and they both reset `next_dentry` to NULL before returning.  The `EventMismatch`, `EmitWarning`, and `NoEvent` labels all exit at the same point.

src/process/path.h
pinealservo: Yeah, it would probably simplify things a bit.

src/file-events.c
pinealservo: I was thinking we'd eventually use it to separate multiple paths in file messages with more than one.  I didn't think much about it and things apparently got re-arranged.  Will delete it and if it turns out we need something like that later we can add it back.

src/file-events.c
pinealservo: The kernel has just checked that the dentry is valid before calling this; I don't think it's worth checking again for a pointer directly passed to us rather than something where we have to find a field via an offset.

src/file-events.c
pinealservo: I'm not sure how it could fail, but I expect it would be obvious from the resulting contents of fm's filter_tag.  Also it will go away if we replace the tag with an integer.

src/file-events.c
pinealservo: I'll check this in case somehow our offsets led to the lookup failing, but this wouldn't have been called by the kernel without a valid path reference.

src/process/path.h
pinealservo: Since path.h is included in both programs via the preprocessor, I have used a preprocessor variable and `#if` blocks to allow programs to choose whether to include the filter stuff or not. It's a little ugly, but ensures the exec programs don't have to deal with the overhead.

rc-andres (approved):

PR #78, opened 2023-06-15
fix mode to be u16 instead of u32

rc-andres:
this was causing us to read too many bytes and the mode came out "wrong" when I was testing it


rc-vbundage (approved):

pinealservo (approved):
Not sure this makes any real difference (both types will hold all valid mode values, and the overall size of the struct will be the same) but it's more clear about how many bits the mode actually uses.

PR #79, opened 2023-06-16
making clangd happy

rc-andres:
This PR should be a no-op. It is mostly just about adding includes on the right spots so `clangd` is happy and added a clangd config because it wasn't happy about our `bpf_helpers.h` file creating variables that itself was not using (but are used by files that import `bpf_helpers.h`.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,15 @@
+---
+If:
+  PathMatch: src/common/bpf_helpers\.h
+Diagnostics:
+  Suppress: -Wunused-variable
+
+---
+CompileFlags:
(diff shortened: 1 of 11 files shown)


src/file/mkdir.h
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+#ifndef USE_PATH_FILTER
+#define USE_PATH_FILTER 1
+#endif
+
```
rc-andres, line 6: alternatively I could add this to `.clangd` for all files inside `src/file` 🤷🏽

pinealservo (approved):

PR #80, opened 2023-06-16
Add more file create event types

pinealservo:
* Moved mkdir.h to create.h
* Finalization code is now in a separate tail-call program, exit_create
* Added probes for symlinks and hard links
* Finalization code can look up source and target paths
* Finalization code will pass filter if either path does
* Can override path lookup loop unroll count


src/file/create.h
```diff
@@ -0,0 +1,169 @@
...
+
+    file_message_t *fm = (file_message_t *)buffer;
+    cursor_t cursor = { .buffer = buffer, .offset = &fm->u.action.buffer_len };
+    int ret = 0;
+    error_info_t info = {0};
+
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    incomplete_create_t *event = bpf_map_lookup_elem(&incomplete_creates, &pid_tgid);
+    if (event == NULL) return;
+    if (event->pid_tgid != pid_tgid) goto EventMismatch;
```
rc-andres, line 100: are we doing this instead of `load_event` to save on instructions? My issue with this is that I no longer have trust in the eBPF maps that the data pointed to by `event` isn't going to be changed in a different probe while this one is running. Granted I've only ever seen this happen in the process events but I never got to the root cause that's why we began to clone the event data and *then* checking that the pid/tgid was correct.

src/file/create.h
```diff
@@ -0,0 +1,169 @@
...
+
+    if (cached_path->next_dentry != NULL) {
+        if (event->target_dentry == NULL) goto ResolveTarget;
```
rc-andres, line 103: I am a bit confused here. We say that if `target_dentry` is `NULL` then resolve the target. but in line 137 we set it to `NULL` to signify we are done with the target so should jump ahead to resolving the source. Should this be the opposite?

src/file/create.h
```diff
@@ -0,0 +1,169 @@
...
+            if (fm->u.action.tag != -1) cached_path->filter_state = -1;
+    }
+
+    ResolveSource:
+    write_path(ctx, cached_path, &cursor, EXIT_CREATE);
+    if (fm->u.action.tag == -1) { 
+        if (cached_path->filter_state <= 0) goto NoEvent;
+        fm->u.action.tag = cached_path->filter_tag;
+    }
+    write_null_char(cursor.buffer, cursor.offset);
```
rc-andres, line 149: I don't think this is necessary since we are done writing paths (although I don't think it hurts)

src/file/create.h
pinealservo: Previously we only accessed the event in the chunk of code that's skipped on tail calls, but we now need to fetch it on every entry.  Furthermore, we'd have to both copy everything out of the map each call, but then copy everything back in when we have to update event->target_dentry to NULL.

If you think it's really a concern, I can switch back to dealing with a copy though.

src/file/create.h
pinealservo: Yeah, I think I may have got the logic reversed on this and didn't notice because I haven't hit a large enough path to tail call yet.

src/file/create.h
rc-andres: Ah I see, and that's because we want to determine where to jump to right? I think in that case we may be overloading what this is meant to do. Maybe we want a new per cpu map to determine that?

If I knew *why* I was getting those weird mismatched events before, or if I had found a bug tracked somewhere then I would feel more strongly about it, right now it's more just a feeling of mistrust without much data on why/when the mismatch occurs.

src/file/create.h
```diff
@@ -95,58 +95,56 @@ static __always_inline void exit_create(struct pt_regs *ctx)
...
-            init_filtered_cached_path(cached_path, event->source, event->target_vfsmount);
+            event.target_dentry = NULL; // mark that we are done with target
+            bpf_map_update_elem(&incomplete_creates, &pid_tgid, &event, BPF_ANY);
+            init_filtered_cached_path(cached_path, event.source, event.target_vfsmount);
             // don't do the filter logic again if we already matched on target
             if (fm->u.action.tag != -1) cached_path->filter_state = -1;
     }
 
     ResolveSource:
+    write_null_char(cursor.buffer, cursor.offset);
```
rc-andres, line 142: we probably want to move this inside the `LINK_HARD` otherwise in cases where we are tail-calling we'll be inserting NULLs inside the same path.

rc-andres (approved):

PR #81, opened 2023-06-21
Delete-like eBPF programs

rc-andres:
This includes unlink, unlinkat, and rmdir probes. 

This is built on top of https://github.com/redcanaryco/redcanary-ebpf-sensor/pull/80 and it's actually very similar so there *may* be some opportunities to refactor but I am not 100% sure how worth it that'd be immediately.

The notable differences are:
* We *never* need a second path so that simplified the exit probe. We may want to modify `MAX_PATH_SEGMENTS_NOTAIL` depending on the type of program. So only defining it in `create.h` or `delete.h` rather than in `file-events.c` although it's also probably not a priority. 
* We need to cache the `dentry->d_inode` during the security call rather than getting it inside the exit probe because by that point the dentry has already become a negative dentry (inode is set to `NULL`). I think this is exactly the opposite as some create ones where if you look at the inode inside the security functions it will be `NULL` as that inode hasn't been created yet.


src/file/create.h
```diff
@@ -140,7 +162,10 @@ static __always_inline void exit_create(struct pt_regs *ctx)
 
     ResolveSource:
     write_null_char(cursor.buffer, cursor.offset);
```
pinealservo, line 164: Don't forget to pick up the fix for this when rebasing/merging.

pinealservo (approved):
Looks good.

FYI, casts to/from void * from/to other pointer types aren't necessary in C, though they don't hurt.  It's C++ that really wants them.  I sometimes use them too, though; not sure why.

PR #83, opened 2023-06-22
Switch mkdir and hardlink to tracepoints

rc-andres:
The goal of this PR is just to make the events that can be a tracepoint be a tracepoint so they can all use the same tail-call for path finding. The only one that's different now is symlink because we want to get the path that the kernel copied out of userspace but that copied path gets deallocated befeore the syscall returns so we gotta hook earlier. We *could* just copy what the user sent instead if we really wanted to use tracepoints for everything but that decision is left for another time.

### DIFF
--- a/src/file/create.h
+++ b/src/file/create.h
@@ -131,7 +131,7 @@ static __always_inline void exit_symlink(struct pt_regs *ctx)
  ResolveTarget:
     ret = write_path(ctx, cached_path, &cursor,
                      (tail_call_t){
-                         .slot = EXIT_CREATE,
+                         .slot = EXIT_SYMLINK,
                          .table = &tail_call_table,
                      });
     if (ret < 0) goto EmitWarning;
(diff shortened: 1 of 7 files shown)


src/file/create.h
```diff
@@ -99,93 +101,118 @@ static __always_inline void exit_create(struct pt_regs *ctx)
...
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    load_event(incomplete_creates, pid_tgid, incomplete_create_t);
+
+    if (event.target_dentry == NULL || event.target_vfsmount == NULL) {
+        set_empty_local_warning(W_NO_DENTRY);
+        goto EmitWarning;
+    }
+
+    // not using read_field_ptr because we expect that `inode` could
+    // be NULL for kernel pseudo filessytems such as cgroupfs. We
```
rc-vbundage, line 181:
```suggestion
    // be NULL for kernel pseudo filesystems such as cgroupfs. We
```

rc-vbundage (approved):

pinealservo (approved):

PR #84, opened 2023-06-23
Chown probes

rc-vbundage:
Adds "chown-like" tracepoints. This just uses what was already setup for the chmod programs which seems to hook in fine. Updated the `Makefile` to be better when working in dev on aarch64. The headers check is deprecated in newer kernel versions.


rc-andres (approved):
Your formatter is so aggressive 😁

PR #88, opened 2023-06-29
emit rename events

rc-andres:
For rename events we had to add some extra logic to the path resolution due to the fact that the kernel modifies the parent + name of the source dentry before the exit of the syscall so we need to grab them before the swap so we can properly get the source path. 

The special rename steps are:
* enter of syscalls: setup an entry in a LRU map for the dentry name, at this point it is zero'd out. We do it here because this probe doesn't do much so we have enough stack space to create the 256 element byte array necessary. 
* security probe: copy the dentry->d_name.name into the spot setup by the enter syscall and save the source dentry parent separately as well.
* exit syscall: setup next_path with the parent dentry
* filemod_paths: if we are resolving something for rename and there is a next path then first copy the saved name. 

Due to stack constraints the name is only loaded from the map as a pointer instead of our paranoid way of loading, copying, and checking. Our thinking is that the errors we had seen in the past may have been exclusive to something going awry with forks/execs and that in filemods that wouldn't be the case.

### DIFF
--- a/src/common/path.h
+++ b/src/common/path.h
@@ -131,6 +131,8 @@ static __always_inline int write_string(const char *string, buf_t *buffer, u32 *
     }
 }
 
+// Writes the single path segment (name) into the buffer. Potentially
+// updating any filter state if `USE_PATH_FILTER` is defined
 static __always_inline int write_segment(void *ctx, cached_path_t *cached_path, cursor_t *buf, void *name) {
 #if USE_PATH_FILTER
     // Path filtering, only if the filter is active
--- a/src/common/path.h
+++ b/src/common/path.h
@@ -131,6 +131,47 @@ static __always_inline int write_string(const char *string, buf_t *buffer, u32 *
     }
 }
 
+static __always_inline int write_segment(void *ctx, cached_path_t *cached_path, cursor_t *buf, void *name) {
+#if USE_PATH_FILTER
+    // Path filtering, only if the filter is active
+    if (cached_path->filter_state >= 0) {
+        filter_key_t key = {0};           // the main key storage for filter transition lookups
+        filter_value_t *val = NULL;
+        key.current_state = cached_path->filter_state;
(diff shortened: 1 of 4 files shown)


src/common/path.h
```diff
@@ -131,6 +131,47 @@ static __always_inline int write_string(const char *string, buf_t *buffer, u32 *
...
+
+    // NAME_MAX doesn't include null character; so +1 to take it
+    // into account. Not all systems enforce NAME_MAX so
+    // truncation may happen per path segment. TODO: emit
+    // truncation metrics to see if we need to care about this.
+    int ret = write_string((char *)name, buf->buffer, buf->offset, NAME_MAX + 1);
+    if (ret < 0) goto WriteError;
+    return 0;
+
+ WriteError:;
```
pinealservo, line 165: You don't need a semicolon after a label unless the syntactic element following isn't a statement, e.g. when you have a variable declaration there.  In C, `if` is a statement rather than an expression, so this should be fine without one.

src/common/types.h
```diff
@@ -405,7 +405,6 @@ typedef struct
             file_ownership_t before_owner;
         } modify;
         struct {
-            file_info_t source;
```
pinealservo, line 408: I guess we don't need this because it's the same info, since the name isn't part of our file_info_t here and we just renamed the file.

pinealservo (approved):
Just an inconsequential nit about the label; otherwise looks great.

pinealservo (approved):

src/common/path.h
```diff
@@ -131,6 +131,47 @@ static __always_inline int write_string(const char *string, buf_t *buffer, u32 *
     }
 }
 
+static __always_inline int write_segment(void *ctx, cached_path_t *cached_path, cursor_t *buf, void *name) {
```
rc-vbundage, line 134: Could we add a doc comment for this function?

rc-vbundage (approved):

PR #89, opened 2023-06-30
Open probes

rc-vbundage:
This PR adds open syscall tracing probes and event generation. Since open can technically be both a create or modify depending on the situation we have to account for it. The `mknod` will go through the modify enter and exit with a different store dentry function (`store_open_create_dentry`) specifically to mark the event as created. This allows the `exit_modify` to then set the event type to `FM_CREATE` or `FM_MODIFY` if the event has the `is_created` flag set to `true`. 

During enter opens we only care if the open was a write open by checking flags and return without saving the map entry if not.


pinealservo (approved):
Very cool; I was a little worried that dealing with create vs. modify might be more complicated, but this handles it very cleanly.

src/file-events.c
```diff
@@ -413,6 +458,149 @@ int BPF_KPROBE(security_path_rename, const struct path *old_dir, struct dentry *
...
+        return 0;
+    exit_modify(ctx);
+    return 0;
+}
+
+
+SEC("tracepoint/sys_enter_creat")
+int tracepoint__syscalls_sys_enter_creat(void *ctx)
+{
+    enter_create(ctx);
```
rc-andres, line 567: creat doesn't always create, it's open and/or create so we need to look at the flags same we do for the other syscalls.

src/file-events.c
```diff
@@ -413,6 +458,149 @@ int BPF_KPROBE(security_path_rename, const struct path *old_dir, struct dentry *
...
+    return 0;
+}
+
+SEC("tracepoint/sys_exit_creat")
+int tracepoint__syscalls_sys_exit__creat(struct syscalls_exit_args *ctx)
+{
+    if (ctx->ret < 0)
+        return 0;
+    exit_create(ctx, LINK_NONE);
```
rc-andres, line 576: same here, `creat` doesn't always create.

src/file-events.c
rc-vbundage: I updated it to call `enter_modify`/`exit_modify`. `creat` doesn't get `flags` as an argument. The syscall just calls `sys_open`, `do_sys_open`, or `ksys_open` depending on kernel version with the flags set to (`O_CREAT | O_WRONLY | O_TRUNCATE`). Can see in [4.14](https://elixir.bootlin.com/linux/v4.14.32/source/fs/open.c#L1115:~:text=i386%20instead%3F%0A%20*/-,SYSCALL_DEFINE2,-(creat%2C) and [6.2](https://elixir.bootlin.com/linux/v6.2.8/source/fs/open.c#:~:text=i386%20instead%3F%0A%20*/-,SYSCALL_DEFINE2,-(creat%2C). So I believe we can just treat it with write flags always.

PR #97, opened 2023-08-25
Use *_kernel or *_user specific functions when reading data

rc-andres:
The initial goal of this PR is use the kernel/user specific functions of `bpf_probe_read` and `bpf_probe_read_str` but since I was already in the area I also touched up the network events a bit but in theory it should be a no-op as I didn't add any new error handling so there is still no warnings being emitted or anything, it is now just more ready to use warnings for whenever we area ready to take that on. 

As far as whether to use the user vs kernel version, it looks like the *only* places where we need the `_user` version is when reading data directly from a user-passed argument in a syscall or when reading the exec args, everything else works with `bpf_probe_read_kernel`. I did check that if I "messed up" and used kernel instead of user or vicecersa then a bunch of our tests began failing so these should be the right ones. 

Newer kernels may not have the _kernel or _user specific functions yet but loaders are smart enough to replace these instructions for the old way on systems where those functions are not defined. 

I recommend looking at this PR in commits.


src/common/common.h
rc-andres, line 1: the changes here are just moving the `helpers.h` file into here since there wasn't really a logic to split of those two files so it made sense to just combine them.

rc-vbundage (approved):

PR #104, opened 2024-02-07
Add a `probe_id` to warning messages to easier debug issues

rc-andres:
`probe_id` is either a syscall number if the probe is a tracepoint; or a custom enum that starts at 0xFFFF with the different kinds of non-tracepoint probes we have for filemod

Additionally, I found a bug in set_current_file_mnt which is used by fchmod where we emit an incorrect warning because we forgot to add a return statement

### DIFF
--- a/src/file/maps.h
+++ b/src/file/maps.h
@@ -9,8 +9,15 @@
 #include "common/types.h"
 #include "file/push_file_message.h"
 
+struct syscalls_enter_generic_args {
+    __u64 unused;
+    long __syscall_nr;
+    /* other args -- make custom structs if needed */
+};
+
 typedef struct {
     file_message_type_t kind;   // kind of file message
+    u64 probe_id;               // ID of the probe that inserted the event
     u64 start_ktime_ns;         // when did the syscall start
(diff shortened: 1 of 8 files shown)


src/file/maps.h
```diff
@@ -131,23 +144,25 @@ static __always_inline incomplete_file_message_t* set_path_mnt(struct pt_regs *c
...
     u64 pid_tgid = bpf_get_current_pid_tgid();
     incomplete_file_message_t *event = bpf_map_lookup_elem(&incomplete_file_messages, &pid_tgid);
     if (event == NULL || event->vfsmount != NULL) return;
 
     void *path = read_field_ptr(file, CRC_FILE_F_PATH);
     if (path == NULL) goto EmitWarning;
-    set_path_mnt(ctx, event, path);
+    set_path_mnt(ctx, event, path, probe_id);
+
+    return;
```
rc-andres, line 161: this missing `return` here was the cause of one of the bugs were we were sending warnings without a warning code. In reality, there was no warning at all to be sent.

rc-vbundage (approved):

pinealservo (approved):

PR #108, opened 2024-07-05
Retrieve btrfs_root->anon_dev as applicable

RomanHargraveRC:
These changes introduce some simple logic to obtain `btrfs_root->anon_dev` for inodes located in a BTRFS volume, following the same logic used by `btrfs_getattr`. 

The device ID found at `struct super_block->s_dev` does not match the value presented to userspace by BTRFS. Instead, BTRFS uses `struct btrfs_root->anon_dev`. Where the sensor had previously only collected `super_block->dev`, it will now collect `btrfs_root->anon_dev` when an inode is located on a BTRFS volume.

### DIFF
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "bpf_helpers.h"
+#include "definitions.h"
 #include "offsets.h"
 #include "types.h"
 #include "warning.h"
@@ -79,6 +80,16 @@ static __always_inline void* ptr_to_field(void *base, u64 field) {
     if (offset == NULL) return NULL;
 
     return base + *offset;
+}
+
+/// Compute the base address of the structure which physically
+/// contains the memory for some pointer. This is simply the address
+/// of the pointer less its offset from the beginning of the structure.
+static __always_inline void *container_of(void *member, u64 field) {
+  u32 *offset = get_offset(field);
+  if (offset == NULL) return NULL;
+
+  return member - *offset;
 }
 
 /// Equivalent to *dst = base->field. Returns 0 on success, or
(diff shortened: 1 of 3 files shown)


src/common/offsets.h
```diff
@@ -155,3 +158,14 @@ const u64 CRC_AUDIT_TASK_INFO_LOGINUID = 0xe6dd107ea26da1a0;
...
+/* BTRFS support */
+
+/* btrfs_inode->root */
+const u64 CRC_BTRFS_INODE_ROOT = 0xb45c091ac97692c3;
+
+/* btrfs_inode->vfs_inode */
+const u64 CRC_BTRFS_INODE_VFS_INODE = 0xc4a24e4c91af5451;
+
+/* btrfs_root->anon_dev */
+const u64 CRC_BTRFS_ROOT_ANON_DEV = 0xa274b01a7473b236;
```
rc-andres, line 171: nit: let's move this above the `LINUX_KERNEL_VERSION` crc because it's an offset like the rest. the `LINUX_KERNEL_VERSION` is more of a hack so it makes sense to be separate.

src/common/common.h
```diff
@@ -185,9 +196,33 @@ static __always_inline int file_info_from_ino(void *inode, file_info_t *file_inf
...
+    // btrfs volume. if it is, the actual device ids that we care
+    // about are embedded in the btrfs_superblock struct.
+    unsigned long s_magic = 0;
+    if (read_field(i_sb, CRC_SBLOCK_S_MAGIC, &s_magic, sizeof(s_magic)) < 0) return -1;
+
+
     // device major/minor
     u32 i_dev = 0;
-    if (read_field(i_sb, CRC_SBLOCK_S_DEV, &i_dev, sizeof(i_dev)) < 0) return -1;
+    if (s_magic != BTRFS_SB_MAGIC) {
```
rc-andres, line 209: nit: since we have an `else` I think I would rather have this `if` be for the case where it is equal (perhaps extracting the how to retrieve the inode into a function to make it shorter?) and the else the "default" case.

rc-andres (approved):

PR #112, opened 2026-03-06
A tracepoint for memfd_create

RomanHargraveRC:
Add a tracepoint and new file message variant to capture details about `memfd_create`. 

For compatibility reasons, the underlying inode and name as the kernel sees it are not retrieved. This is something we can improve with CORE.

### DIFF
--- a/src/file-events.c
+++ b/src/file-events.c
@@ -15,6 +15,7 @@
 #include "file/create.h"
 #include "file/delete.h"
 #include "file/maps.h"
+#include "file/memfd.h"
 #include "file/modify.h"
 #include "file/rename.h"
 
@@ -536,6 +537,21 @@ int BPF_KPROBE(security_inode_rename, struct inode *old_dir, struct dentry *old_
 
 /* END RENAME PROBES */
 /* BEGIN OPEN-LIKE PROBES */
+
+SEC("tracepoint/syscalls/sys_enter_memfd_create")
+int sys_enter_memfd_create(struct syscalls_enter_memfd_create_args *ctx)
+{
+    enter_memfd_create(ctx);
+    return 0;
+}
+
+SEC("tracepoint/syscalls/sys_exit_memfd_create")
+int sys_exit_memfd_create(struct syscalls_exit_args *ctx)
+{
+    // exit_memfd pushes the message
+    (void) POP_AND_SETUP(FM_MEMFD_CREATE, exit_memfd);
+    return 0;
+}
 
 SEC("tracepoint/syscalls/sys_enter_open")
 int sys_enter_open(struct syscalls_enter_open_args *ctx)
--- a/src/file-events.c
+++ b/src/file-events.c
@@ -76,7 +76,7 @@ int BPF_KPROBE(security_inode_mkdir, const struct inode *dir, struct dentry *den
 SEC("tracepoint/syscalls/sys_exit_mkdir")
 int sys_exit_mkdir(struct syscalls_exit_args *ctx)
 {
-    file_message_t *fm = POP_AND_SETUP_ARGS(FM_CREATE, exit_create, LINK_NONE);
+    file_message_t *fm = POP_AND_THEN_ARGS(FM_CREATE, exit_create, LINK_NONE);
     finish_message(ctx, fm);
     return 0;
 }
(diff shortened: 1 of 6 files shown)


rc-andres (commented):
-- IGNORE --

src/file/memfd.h
```diff
@@ -0,0 +1,77 @@
...
+            fm.u.warning.info.err = read_len;
+
+            push_file_message((struct syscalls_enter_generic_args *)ctx, &fm);
+
+            // this should not be fatal. the name is informative, but has no
+            // function. ensure that there isn't an unterminated string in the
+            // buffer as the result of the failed read.
+            name_buf[0] = (char)0;
+
+            read_len = 0;
```
rc-andres, line 74: should we return at this point to prevent the `push_flexible_file_message` from firing?

src/file/memfd.h
rc-andres, line 1: should we add #pragma once like we do in other files?

src/file/memfd.h
RomanHargraveRC: My intention was that we would simply not capture the memfd name on a failed read. With the added prefix logic, it may be desirable to remove that as well; however, so long as the syscall itself succeeds we should record the event, I would imagine.

src/file/memfd.h
RomanHargraveRC: It would be fine to do that, although this file should not be included multiple times (and many inclusions could point to some sloppy code) due to its function being more analogous to a source unit in a traditional C application than as a header.

src/file/maps.h
```diff
@@ -55,25 +59,29 @@ struct bpf_map_def SEC("maps/incomplete_file_messages") incomplete_file_messages
     .namespace = "",
 };
 
-static __always_inline void enter_file_message(struct syscalls_enter_generic_args *ctx, file_message_type_t kind)
+static __always_inline void prepare_file_message(struct syscalls_enter_generic_args* ctx, file_message_type_t kind, incomplete_file_message_t* event)
+{
+    *event = (incomplete_file_message_t){0};
```
rc-andres, line 64: isn't this already zero'd out on callers?

src/file-events.c
```diff
@@ -537,6 +538,21 @@ int BPF_KPROBE(security_inode_rename, struct inode *old_dir, struct dentry *old_
...
+{
+    enter_memfd_create(ctx);
+    return 0;
+}
+
+SEC("tracepoint/syscalls/sys_exit_memfd_create")
+int sys_exit_memfd_create(struct syscalls_exit_args *ctx)
+{
+    // exit_memfd pushes the message
+    (void) POP_AND_SETUP(FM_MEMFD_CREATE, exit_memfd);
```
rc-andres: we discussed renaming this to `POP_AND_THEN` or `POP_AND_DO`. Is this still something we want to do?

src/file/maps.h
RomanHargraveRC: It is, though it's nice to be explicit. I'd be fine with removing it.

src/file-events.c
RomanHargraveRC: easy enough.

src/file/memfd.h
rc-andres: I agree probably not necessary since it isn't a real header but for the sake of avoiding any weirdness later on if it did became a header let's just add it

src/file/memfd.h
```diff
@@ -0,0 +1,77 @@
+#include "common/bpf_helpers.h"
+#include "common/types.h"
+#include "maps.h"
+#include "modify.h"
```
rc-andres: why are we we including `modify.h` here?

src/file/memfd.h
```diff
@@ -0,0 +1,77 @@
...
+#include "push_file_message.h"
+#include "vmlinux.h"
+
+#include "common/definitions.h"
+
+static const char MEMFD_PREFIX[] = "memfd:";
+
+#define MEMFD_PREFIX_LEN sizeof(MEMFD_PREFIX) - 1
```
rc-andres: parens around this just to make any math using the macro doesn't mess up by accident

src/file/memfd.h
```diff
@@ -0,0 +1,77 @@
...
+const unsigned long MEMFD_NAME_MAX = NAME_MAX - MEMFD_PREFIX_LEN;
```
rc-andres: static const to be the same as the other ones? I am assuming we did that for a reason (older kernels?)

src/file/memfd.h
rc-andres: In fact the math below is incorrect due to the missing parens I am pretty sure

src/file/memfd.h
RomanHargraveRC: I had grouped `syscalls_enter_memfd_create_args` alongside the similar `open*` syscall structures. Why, I do not know. 

clangd will have added this as part of a CAP insertion.

src/file/memfd.h
```diff
@@ -0,0 +1,84 @@
...
+#include "push_file_message.h"
+#include "vmlinux.h"
+
+#include "common/definitions.h"
+
+struct syscalls_enter_memfd_create_args {
+   __u64 unused;
+   long __syscall_nr;
+   const char *uname;
+   unsigned long flags;
```
rc-andres: let's add one more space to make it consistent with our other 4-space identiation

rc-andres (approved):

